PR #7, opened 2023-06-23
fix: describe to work with parameterized SQL queries

criccomini:
I ran into an issue when using fakesnow with parameterized SQL queries. I was seeing:

```
    def description(self) -> list[ResultMetadata]:
        # use a cursor to avoid destroying an unfetched result on the main connection
        with self._duck_conn.cursor() as cur:
            assert self._conn.database, "Not implemented when database is None"
            assert self._conn.schema, "Not implemented when database is None"

            # match database and schema used on the main connection
            cur.execute(f"SET SCHEMA = '{self._conn.database}.{self._conn.schema}'")
>           cur.execute(f"DESCRIBE {self._last_sql}")
E           duckdb.InvalidInputException: Invalid Input Error: Prepared statement needs 1 parameters, 0 given
```

The issue was that DESCRIBE was assuming _last_sql parameters were already filled in, but this isn't the case for server-side parameter binding. See:

https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-example#qmark-or-numeric-binding

For some notes on this.

I found that saving the params like the SQL string fixed the issue. I've added a new test that failed before, but passes now.

### DIFF
--- a/fakesnow/fakes.py
+++ b/fakesnow/fakes.py
@@ -57,6 +57,7 @@ class FakeSnowflakeCursor:
         self._duck_conn = duck_conn
         self._use_dict_result = use_dict_result
         self._last_sql = None
+        self._last_params = None
 
     def __enter__(self) -> Self:
         return self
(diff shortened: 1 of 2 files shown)


tekumara:
Nice catch, thanks for the fix!

Issue #43, opened 2024-02-09
Support for `SHOW USERS`


DanCardin:
I wonder if it would be possible/practical to have provisional support for CRUD on roles/users? I was sort of imagining part of setup would create an `information_schema.fakesnow_users` table (since there's no information_schema table that i can see, unlike other SHOW commands that **do** have public backing tables) that matches the structure given by `SHOW USERS`.

Then `SHOW USERS` could function, even if it only returned an empty table.

But at that point, basic `CREATE/UPDATE/DELETE` `USER/ROLE` statements could probably be supported by inserting into that table.

---
If you'd support such an endeavor, I'd be interested in supplying the impl.

I believe I see that sqlglot would need to gain `SHOW USERS` syntactically. At which point I could implement up to `SHOW USERS`. Beyond that, my usecase could probably live with manually manipulating `fakesnow_users`, although certainly it'd be nice to keep going.

---

fwiw, my usecase is for adding snowflake support to https://github.com/DanCardin/sqlalchemy-declarative-extensions, wherein my usual e2e testing mechanism for the rest of the library is impractical for snowflake. Developing against a very rough approximation of users isn't ideal, but it seems better than the manual patching i would otherwise be forced to do.

tekumara:
Yes this is very doable and, like any unimplemented feature, a welcome addition! 🙂 

If you wanted to have a go, there's some precedence for storing extra metadata per-database in `_fs_` prefixed tables, which might be a useful reference, see here: https://github.com/tekumara/fakesnow/blob/a86f238accc4d3957b684e8e96cb31153918d994/fakesnow/info_schema.py#L7 

The difference would be that a users table isn't per database but per session. So I'd suggest creating a [SNOWFLAKE database](https://docs.snowflake.com/en/sql-reference/snowflake-db) and storing the users table under a `_fs_` prefix there.

Hopefully there's enough to get started in [CONTRIBUTING.md](https://github.com/tekumara/fakesnow/blob/main/CONTRIBUTING.md). Any questions let me know and I'd be happy to help.

PS: nice use-case and library!

tekumara:
BTW - if you only want to support `SHOW USERS` returning an empty result set, then you could do so without a table, and just transform `SHOW USERS` into a SELECT with 0 rows.

[...]

PR #47, opened 2024-02-13
feat: Implement SHOW USERS.

DanCardin:
Fixes https://github.com/tekumara/fakesnow/issues/43

* I **did** have to bump to 21.1.0. I can certainly supply that as a separate PR, but I didn't have to change anything to make your existing tests pass, and the changelog for 21.0.0 didnt include anything that seemed like it applied to this lib.
* I relocated all the on-connection-creation setup code into a `connection.py`, because it seemed like it could just end up growing unboundedly, and it was fairly distinct from the rest of the code in `fakes.py`. I can certainly put it back otherwise.
* Similarly, I put the tests for the new functionality in a separate module. Seemed like it might be easier to follow features by file versus trying to slot it into the end of `test_fakes.py`, but again, can just put it where you want.


fakesnow/connection.py
```diff
@@ -0,0 +1,125 @@
...
+    return database_set, schema_set
+
+
+def create_global_database(conn: DuckDBPyConnection) -> None:
+    """Create a "global" database for storing objects which span database.
+
+    Including (but not limited to):
+     - Users
+    """
+    with contextlib.suppress(BinderException):
```
DanCardin, line 122: I guess there is a way of listing attached databases, but implementing it would have required a decent amount more code and I wasn't sure it would have been worth it

tekumara: Does `ATTACH IF NOT EXISTS` work here instead?

pyproject.toml
```diff
@@ -11,7 +11,7 @@ dependencies = [
     "duckdb~=0.9.2",
     "pyarrow",
     "snowflake-connector-python",
-    "sqlglot~=21.0.1",
+    "sqlglot~=21.1.0",
```
tekumara, line 14: 👍

tests/test_fakes.py
```diff
@@ -798,7 +799,10 @@ def test_info_schema_databases(conn: snowflake.connector.SnowflakeConnection):
         cur.execute("create database db2")
         cur.execute("select * from information_schema.databases")
 
-        assert cur.fetchall() == [
+        databases = typing.cast(list[dict], cur.fetchall())
+        result = [d for d in databases if d["database_name"] != "_fs_global"]
```
tekumara, line 803: Can we filter this out in the implementation, see [SQL_CREATE_INFORMATION_SCHEMA_DATABASES_VIEW ](https://github.com/tekumara/fakesnow/blob/b8e38a6ecf28173c11713a8ab5ca6287c8114ebb/fakesnow/info_schema.py#L64), rather than expose this implementation detail to end users?

tests/test_show_users.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def test_show_users_base_case(cur: snowflake.connector.cursor.SnowflakeCursor):
+    result = cur.execute("SHOW USERS")
+    assert result
+    assert result.fetchall() == []
+
+
+def test_show_users_with_users(cur: snowflake.connector.cursor.SnowflakeCursor):
+    cur.execute(f"INSERT INTO {USERS_TABLE_FQ_NAME} (name) VALUES ('foo'), ('bar')")
```
tekumara, line 13: One of goals of the tests is that they pass on a real Snowflake instance (see [CONTRIBUTING.MD](https://github.com/tekumara/fakesnow/blob/main/CONTRIBUTING.md#raising-a-pr)). The tests also document the behaviour end users can expect will be stable, and I'd prefer not to guarantee an implementation detail like an internal database or table name won't change.

Ideally this would use `CREATE USER` which would satisfy the criteria above. But failing that, could we remove this test?

tests/test_show_users.py
```diff
@@ -0,0 +1,20 @@
+import snowflake.connector.cursor
```
tekumara, line 1: I don't mind splitting this out but can I suggest:
1. Updating the README.md which currently only points at test_fakes.py
1. Having a coarser grain. In the limit, if we end up with a file per command, that'll be a lot of files and potentially hard to navigate. _test_show.py_ containing all the SHOW functionality I think would be fine.

fakesnow/connection.py
DanCardin: oooo. I was looking at sqlite docs, where that's not a thing. but duckdb!

DanCardin: although this syntax is only available in duckdb 0.10, where you're currently bound to 0.9.2. just fwiw

tests/test_show_users.py
DanCardin: It seemed important to have a test that asserted...anything about the tables existing columns or otherwise the result not being empty.

tests/test_show_users.py
DanCardin: fwiw In the existing limit, I found the majority of tests in test_fakes and test_transforms to be hard to navigate. My personal preference would be to only colocate things that are likely to fail together...but it's not my library.

I was thinking perhaps `test_users.py`, which would end up colocating tests for `SHOW USERS` which will inevitably also end up inside any tests for `CREATE USER`

fakesnow/connection.py
```diff
@@ -0,0 +1,125 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
DanCardin: So i did that, but it introduced a circular import issue because transforms need to import the database/table name.

Thus I created a `global_database.py` instead, which leaves `connection.py` with a largely unaltered `FakeSnowflakeConnection`. I can just put that back in `fakes.py` if you want, or leave it here.

DanCardin:
Pushed an update that does:
* `ATTACH IF NOT EXISTS` and bumps duckdb to 0.10 minimum
* implements the most basic `CREATE USER` statement I could implement for tests. my further enhancing sqlglot is the blocker for this to become more sophisticated
* I put back the connction code and made `global_database.py` to encapsulate stuff that goes in there

tekumara:
Thank you, looks great! I pushed a commit to tidy up the diff and, to address the change introduced in duckdb 0.10.0, I fixed the implementation of `test_flatten` rather than change the test, to preserve the existing behaviour. 

It's now been released:  https://github.com/tekumara/fakesnow/releases/tag/v0.9.2

PR #50, opened 2024-02-17
feat: implement information_schema.views

DanCardin:
In the 2nd commit, I relocate all the information_schema-related tests into a single file (again, somewhat opinionated decision, but infinitely adding tests to test_fakes seemed like perhaps not ideal, as the info schema stuff is only likely to get larger).

### DIFF
--- a/tests/test_info_schema.py
+++ b/tests/test_info_schema.py
@@ -1,36 +1,194 @@
-from datetime import datetime, timezone
+# ruff: noqa: E501
 
+from datetime import datetime
+
+import pytz
 import snowflake.connector.cursor
+from snowflake.connector.cursor import ResultMetadata
 
-UTC = timezone.utc
+
+def test_info_schema_table_comments(cur: snowflake.connector.cursor.SnowflakeCursor):
+    def read_comment() -> str:
+        cur.execute(
+            """SELECT COALESCE(COMMENT, '') FROM INFORMATION_SCHEMA.TABLES
+                    WHERE TABLE_NAME = 'INGREDIENTS' AND TABLE_SCHEMA = 'SCHEMA1' LIMIT 1"""
+        )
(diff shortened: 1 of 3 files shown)


tests/test_info_schema.py
```diff
@@ -0,0 +1,193 @@
...
+    assert result.fetchall() == []
+
+
+def test_info_schema_views_with_views(cur: snowflake.connector.cursor.SnowflakeCursor):
+    cur.execute("CREATE TABLE foo (id INTEGER, name VARCHAR)")
+    cur.execute("CREATE VIEW bar (SELECT * FROM foo WHERE id > 5)")
+
+    result = cur.execute("SELECT * FROM information_schema.views")
+    assert result
+    assert result.fetchall() == [
```
tekumara, line 176: could we use a `DictCursor` here to verify the column names too? (see `test_info_schema_databases` for an example)

tests/test_info_schema.py
```diff
@@ -0,0 +1,193 @@
...
+
+def test_info_schema_views_empty(cur: snowflake.connector.cursor.SnowflakeCursor):
+    result = cur.execute("SELECT * FROM information_schema.views")
+    assert result
+    assert result.fetchall() == []
+
+
+def test_info_schema_views_with_views(cur: snowflake.connector.cursor.SnowflakeCursor):
+    cur.execute("CREATE TABLE foo (id INTEGER, name VARCHAR)")
+    cur.execute("CREATE VIEW bar (SELECT * FROM foo WHERE id > 5)")
```
tekumara, line 172: to make this work on a real Snowflake instance:
```suggestion
    cur.execute("CREATE VIEW bar AS SELECT * FROM foo WHERE id > 5")
```

tekumara (commented):
Thank you! This is a nice addition.

PR #55, opened 2024-02-26
feat: support sqlalchemy metadata reflection

DanCardin:
The main drawback: I dont see a way to rigorously support the `pk_` columns of `IMPORTED KEYS`. And/or generally a bunch of the snowflake columns, I am hardcoding values for.

From a full compatibility perspective, it's not great. But realistically, it will never be perfect, and it **seems** to be irrelevant to SQLAlchemy


tests/test_fakes.py
```diff
@@ -875,13 +875,15 @@ def test_show_objects(dcur: snowflake.connector.cursor.SnowflakeCursor):
             "kind": "TABLE",
             "database_name": "DB1",
             "schema_name": "SCHEMA1",
+            "comment": None,
```
tekumara, line 878: snowflake doesn't show the `comment` field for a **terse** summary. i'd like to make sure terse doesn't have extra columns that someone might unexpectedly come to depend on.

`comment` appears as part of `show objects` which has this complete set of fields:

created_on, name, database_name, schema_name, kind, comment, cluster_by, rows, bytes, owner, retention_time, owner_role_type, budget

tests/test_fakes.py
```diff
@@ -938,12 +949,20 @@ def test_show_tables(dcur: snowflake.connector.cursor.SnowflakeCursor):
...
         },
     ]
     # assert dcur.fetchall() == objects
     dcur.execute("show terse tables in db1.schema1")
     assert dcur.fetchall() == objects
-    assert [r.name for r in dcur.description] == ["created_on", "name", "kind", "database_name", "schema_name"]
+    assert [r.name.lower() for r in dcur.description] == [
```
tekumara, line 958: snowflake will return these column names as lower case, but here `COMMENT` is appearing as uppercase

to make it `comment` lowercase, you can quote it in the query, eg: `NULL as "comment"`

tekumara (commented):
Thank you for this. Hardcoding values is fine, I've made some suggestion below though to ensure the column names are correct.

DanCardin:
Good callouts. Added a 2nd commit which makes that distinction.

tests/test_fakes.py
```diff
@@ -938,12 +947,20 @@ def test_show_tables(dcur: snowflake.connector.cursor.SnowflakeCursor):
             "kind": "TABLE",
             "database_name": "DB1",
             "schema_name": "SCHEMA1",
+            "comment": None,
```
tekumara, line 950: similar to objects, `terse schemas` doesn't have the comment column
```suggestion
```

tests/test_fakes.py
```diff
@@ -884,7 +884,6 @@ def test_show_objects(dcur: snowflake.connector.cursor.SnowflakeCursor):
             "schema_name": "SCHEMA1",
         },
     ]
-    assert dcur.fetchall() == objects
```
tekumara, line 887: this should still work i think?

tests/test_fakes.py
```diff
@@ -938,12 +947,20 @@ def test_show_tables(dcur: snowflake.connector.cursor.SnowflakeCursor):
...
     dcur.execute("show terse tables in db1.schema1")
     assert dcur.fetchall() == objects
-    assert [r.name for r in dcur.description] == ["created_on", "name", "kind", "database_name", "schema_name"]
+    assert [r.name.lower() for r in dcur.description] == [
+        "created_on",
+        "name",
+        "kind",
+        "database_name",
+        "schema_name",
+        "comment",
```
tekumara, line 962: likewise here, `terse tables` doesn't have the comment column
```suggestion
    assert [r.name for r in dcur.description] == ["created_on", "name", "kind", "database_name", "schema_name"]
```

tekumara (commented):
thanks for the updates, the same applies to the other `show` kinds, see below

tests/test_fakes.py
DanCardin: gotcha. I know for all three of the new ones, `terse` has no effect

tests/test_fakes.py
DanCardin: It's just that i'm testing both terse and non-terse, so I'd have to mutate the original structure to add the "comment
" column. i figured it was less relevant than just showing the column now exists

tekumara:
Thank you @DanCardin! I've pushed some TODOs and an `xfail` test to document whats needed for full SHOW KEYS support in future work.

PR #62, opened 2024-03-26
refactor: remove parse_json as handled by sqlglot

seruman:
Seems like already handled by `sqlglot`, removing transformation won't break any tests;
```python
q1 = """SELECT PARSE_JSON('{"k1": "v1"}')"""
parse_one(q1, dialect="snowflake").sql(dialect="duckdb")

q2 = """SELECT PARSE_JSON(col)"""
parse_one(q2, dialect="snowflake").sql(dialect="duckdb")
```

```python
Select(
  expressions=[
    ParseJSON(
      this=Literal(this={"k1": "v1"}, is_string=True))])
# SELECT JSON('{"k1": "v1"}')

Select(
  expressions=[
    ParseJSON(
      this=Column(
        this=Identifier(this=col, quoted=False)))])
# SELECT JSON(col)
```

----

Left `test_parse_json` as is without any transformation in case of any `sqlglot` changes. There're other tests those cover it in fakes tests though deleting kinda makes sense.

------------------

* Python: 3.9.18
* sqlglot: 21.2.1

### DIFF
--- a/tests/test_transforms.py
+++ b/tests/test_transforms.py
@@ -23,7 +23,6 @@ from fakesnow.transforms import (
     json_extract_cast_as_varchar,
     json_extract_precedence,
     object_construct,
-    parse_json,
     random,
     regex_replace,
     regex_substr,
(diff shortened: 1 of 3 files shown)


tests/test_transforms.py
```diff
@@ -264,9 +263,10 @@ def test_object_construct() -> None:
...
     assert (
-        sqlglot.parse_one("""insert into table1 (name) select parse_json('{"first":"foo", "last":"bar"}')""")
-        .transform(parse_json)
-        .sql(dialect="duckdb")
+        sqlglot.parse_one(
+            """insert into table1 (name) select parse_json('{"first":"foo", "last":"bar"}')""",
+            read="snowflake",
+        ).sql(dialect="duckdb")
         == """INSERT INTO table1 (name) SELECT JSON('{"first":"foo", "last":"bar"}')"""
     )
```
tekumara, line 272: may as well delete the whole test!
```suggestion
```

tekumara (commented):
good spot, thanks for this!

tests/test_transforms.py
seruman: deleted @https://github.com/tekumara/fakesnow/pull/62/commits/9929d0c3638851829acfe741c5bfea97733571ec

PR #63, opened 2024-03-27
feat: support sha2 and sha2_hex with digest size of 256

ifm-pgarner:
We're using Snowflake `sha2` function in our project that uses `fakesnow` for testing (it's great, thank you!)

But duckdb does not have `sha2` function, instead it has a `sha256` function

`sha256(foo)` is the same as Snowflake `sha2('foo', 256)` (or you can omit the length arg and it defaults to 256) https://docs.snowflake.com/en/sql-reference/functions/sha2#arguments

So this PR adds a transform for these function calls

### DIFF
--- a/tests/test_transforms.py
+++ b/tests/test_transforms.py
@@ -634,3 +634,34 @@ def test_sha256() -> None:
         sqlglot.parse_one("insert into table1 (name) select sha2_hex('foo', 256, 'wtf')").transform(sha256).sql()
         == "INSERT INTO table1 (name) SELECT SHA2_HEX('foo', 256, 'wtf')"
     )
+
+
+def test_sha256_binary() -> None:
+    # snowflake default sha2 length is 256
+    assert (
+        sqlglot.parse_one("insert into table1 (name) select sha2_binary('foo')").transform(sha256).sql(dialect="duckdb")
+        == "INSERT INTO table1 (name) SELECT UNHEX(SHA256('foo'))"
+    )
+
+    assert (
+        sqlglot.parse_one("insert into table1 (name) select sha2_binary('foo', 256)")
+        .transform(sha256)
+        .sql(dialect="duckdb")
+        == "INSERT INTO table1 (name) SELECT UNHEX(SHA256('foo'))"
+    )
+
+    # values with hash length other than 256 are not transformed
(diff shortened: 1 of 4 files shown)


tests/test_transforms.py
```diff
@@ -424,3 +425,48 @@ def test_values_columns() -> None:
...
+    assert (
+        sqlglot.parse_one("insert into table1 (name) select sha2_hex()").transform(values_columns).sql()
+        == "INSERT INTO table1 (name) SELECT SHA2_HEX()"
+    )
+    assert (
+        sqlglot.parse_one("insert into table1 (name) select sha2_hex('foo', 256, 'wtf')")
+        .transform(values_columns)
+        .sql()
+        == "INSERT INTO table1 (name) SELECT SHA2_HEX('foo', 256, 'wtf')"
+    )
```
tekumara, line 472: would you mind also adding a test to test_fakes.py too? this will guarantee integrated test coverage that includes duckdb's execution

tekumara (commented):
looks great, thank you for this!

anentropic:
I realised I could use `sha2_binary` and store my hashes in a 32 byte column instead of varchar(64) for hexdigest

duckdb doesn't have that function but `unhex(sha256(...))` seems to give same result

```python
In [24]: duckdb.sql("SELECT unhex(sha256('Snowflake'))").show()
┌────────────────────────────────────────────────────────────────────────────────────────┐
│                               unhex(sha256('Snowflake'))                               │
│                                          blob                                          │
├────────────────────────────────────────────────────────────────────────────────────────┤
│ \x1D\xBDY\xF6a\xD6\x8B\x90rO!\x08C\x96\xB8eIqs\xE4\xD2qOM\x91\xCF\x05\xFA_\xC5\xE1\x8D │
└────────────────────────────────────────────────────────────────────────────────────────┘


In [25]: sha256(b"Snowflake").digest()
Out[25]: b'\x1d\xbdY\xf6a\xd6\x8b\x90rO!\x08C\x96\xb8eIqs\xe4\xd2qOM\x91\xcf\x05\xfa_\xc5\xe1\x8d'
```

so I have added that transform as well

tekumara:
Perfect, thank you!

PR #65, opened 2024-03-27
fix: write_pandas quotes columns when inserting dataframe

ifm-pgarner:
`_insert_df` needs to quote the column names when inserting, otherwise fails for table where column names need quoting


tekumara:
Thank you! I extended the test to include checking the value of the column names returned

PR #66, opened 2024-03-27
feat: add ARRAY_AGG() WITHIN GROUP() support

seruman:
In Snowflake, to order elements in `ARRAY_AGG` aggregation one needs to provide `WITHIN GROUP (..)`;
* https://docs.snowflake.com/en/sql-reference/functions/array_agg#arguments
    
    ```
    ARRAY_AGG(<expr>) WITHIN GROUP (<order-by-clause>)
    ```
In DuckDB, `LIST()/ARRAY_AGG()` requires expression to be passed ordered and does not support `WITHIN GROUP (...)`;
* https://duckdb.org/docs/sql/aggregates.html#order-by-clause-in-aggregate-functions
    ```
    ARRAY_AGG( <expr> <order-by-clause> )
    ```

This transformer simply combines *aggreage* expression in `ARRAY_AGG` and *ordering* one in `WITHIN GROUP` in an *order* expression.

```sql
ARRAY_AGG(DISTINCT id) WITHIN GROUP (ORDER BY id)
```

```python
WithinGroup(
  this=ArrayAgg(
    this=Distinct(
      expressions=[
        Column(
          this=Identifier(this=id, quoted=False))])),
  expression=Order(
    expressions=[
      Ordered(
        this=Column(
          this=Identifier(this=id, quoted=False)),
        nulls_first=False)]))
```

```python
ArrayAgg(
  this=Order(
    this=Distinct(
      expressions=[
        Column(
          this=Identifier(this=ID, quoted=False))]),
    expressions=[
      Ordered(
        this=Column(
          this=Identifier(this=ID, quoted=False)),
        nulls_first=False)]))
```

```diff
-WithinGroup(
-  this=ArrayAgg(
+ArrayAgg(
+  this=Order(
     this=Distinct(
       expressions=[
         Column(
-          this=Identifier(this=id, quoted=False))])),
-  expression=Order(
+          this=Identifier(this=ID, quoted=False))]),
     expressions=[
       Ordered(
         this=Column(
-          this=Identifier(this=id, quoted=False)),
+          this=Identifier(this=ID, quoted=False)),
         nulls_first=False)]))

```

-----

Snowflake has following limitation though;
> https://docs.snowflake.com/en/sql-reference/functions/array_agg#usage-notes
> If you specify DISTINCT and WITHIN GROUP, both must refer to the same column. For example:

I've left it out of the scope of this PR to keep it simple.

### DIFF
--- a/fakesnow/transforms.py
+++ b/fakesnow/transforms.py
@@ -22,6 +22,32 @@ def array_size(expression: exp.Expression) -> exp.Expression:
     return expression
 
 
+def array_agg_within_group(expression: exp.Expression) -> exp.Expression:
+    """Convert ARRAY_AGG(<expr>) WITHIN GROUP (<order-by-clause>) to ARRAY_AGG( <expr> <order-by-clause> )
+    Snowflake uses ARRAY_AGG(<expr>) WITHIN GROUP (ORDER BY <order-by-clause>)
+    to order the array, but DuckDB uses ARRAY_AGG( <expr> <order-by-clause> ).
+    See;
+        - https://docs.snowflake.com/en/sql-reference/functions/array_agg
+        - https://duckdb.org/docs/sql/aggregates.html#order-by-clause-in-aggregate-functions
+    Note; Snowflake has following restriction;
+            If you specify DISTINCT and WITHIN GROUP, both must refer to the same column.
+          Transformation does not handle this restriction.
+    """
+    if (
+        isinstance(expression, exp.WithinGroup)
+        and (agg := expression.find(exp.ArrayAgg))
+        and (order := expression.expression)
+    ):
+        return exp.ArrayAgg(
+            this=exp.Order(
+                this=agg.this,
+                expressions=order.expressions,
+            )
+        )
+
+    return expression
+
+
 # TODO: move this into a Dialect as a transpilation
 def create_database(expression: exp.Expression, db_path: Path | None = None) -> exp.Expression:
     """Transform create database to attach database.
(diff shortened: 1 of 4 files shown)


fakesnow/transforms.py
```diff
@@ -22,6 +22,32 @@ def array_size(expression: exp.Expression) -> exp.Expression:
...
+        isinstance(expression, exp.WithinGroup)
+        and (agg := expression.find(exp.ArrayAgg))
+        and (order := expression.expression)
+    ):
+        return exp.ArrayAgg(
+            this=exp.Order(
+                this=agg.this,
+                expressions=order.expressions,
+            )
+        )
```
tekumara, line 46: to produce a json array
```suggestion
        return exp.Anonymous(this="TO_JSON", expressions=[exp.ArrayAgg(
            this=exp.Order(
                this=agg.this,
                expressions=order.expressions,
            )
        )])
 ```

tekumara (commented):
👍

fakesnow/transforms.py
seruman: missed this. besides this case, all `ArrayAgg`'s might need to be wrapped in `TO_JSON`. 

how about adding another transform that runs after this one that wraps `ArrayAgg`s?

seruman: ☝️ what i meant was something like this;
https://gist.github.com/seruman/79b050a0ff96c90d616350a2ca62dfaf

tekumara: Ah yep!

seruman: thanks you, updated @https://github.com/tekumara/fakesnow/pull/66/commits/3c310cc14ac317698e4e19e404ea71c13c387e98

tekumara:
Thank you for this!

FYI I pushed some small changes in d50873a so the tests would pass on the real snowflake, and made some used the `dcur` fixture to simplify the test.

PR #67, opened 2024-03-27
feat: add TRY_PARSE_JSON

seruman:
Pretty similar to -former- `parse_json` transform; wraps expression in `TRY_CAST(<expr> as JSON)`.


tekumara:
Thank you!

PR #68, opened 2024-03-27
feat: add TRY_TO_{DECIMAL,...}

seruman:
Adds transformation for `TRY_TO_DECIMAL` and friends, basically the same as `to_decimal` except returned node is `TryCast` instead of `Cast`.

Moved the actual node creation to another function to reuse it in both `*to_decimal` transformation, but not 100% sure with what I end up with. 

Keeping the same logic in both functions makes it more explicit and more isolated but prone to failures in a case that is not covered by tests, would like to hear what you think about it.

----------------
> [!NOTE]  
> I'm working on a wrapper dbt adapter so users can write unit tests and run locally. Thank you for this awesome project as it really sped things up and kept my focus solely on query transformations.
>
> I have a couple more transformation but not sure if all fits in the scope of the project as some are quite opinionated.  I wanted to open a separate PR for each transformation to keep it reviewable but it might get annoying for a maintainer. Please let me know your thoughts, like if you prefer them to be in a single PR.

### DIFF
--- a/fakesnow/fakes.py
+++ b/fakesnow/fakes.py
@@ -185,6 +185,7 @@ class FakeSnowflakeCursor:
             .transform(transforms.values_columns)
             .transform(transforms.to_date)
             .transform(transforms.to_decimal)
+            .transform(transforms.try_to_decimal)
             .transform(transforms.to_timestamp_ntz)
             .transform(transforms.to_timestamp)
             .transform(transforms.object_construct)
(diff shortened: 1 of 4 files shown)


tekumara:
Looks good to me! 

And thank you for splitting each transformation into its own separate PR, it makes it a lot easier for me to review. Plus, because the changelog is generated automatically from merged commits, it ends up being much cleaner.

PR #69, opened 2024-03-28
feat: support trim with numeric and variant types

seruman:
Snowflake's [`TRIM`](https://docs.snowflake.com/en/sql-reference/functions/trim) accepts any type and returns VARCHAR, docs does not mentioned it but it seems to only works on VARCHAR/TEXT typed inputs and acts as a cast to VARCHAR for others.  DuckDB's `TRIM` only accepts VARCHAR. It seemed safe to me to cast input to VARCHAR to mimic similar behaviour.
```sql
select 
    trim(5) as c_number, SYSTEM$TYPEOF(c_number),
    trim(' str ') as c_varchar, SYSTEM$TYPEOF(c_varchar),
    trim(to_variant({'k': 'v'})) as c_variant, SYSTEM$TYPEOF(c_variant)
;
```

| C_NUMBER | SYSTEM$TYPEOF(C_NUMBER) | C_VARCHAR | SYSTEM$TYPEOF(C_VARCHAR) | C_VARIANT | SYSTEM$TYPEOF(C_VARIANT) |
|----------|-------------------------|-----------|--------------------------|-----------|--------------------------|
|        5 | VARCHAR(16777216)[LOB]  | str       | VARCHAR(5)[LOB]          | {"k":"v"} | VARCHAR(16777216)[LOB]   |

### DIFF
--- a/tests/test_fakes.py
+++ b/tests/test_fakes.py
@@ -1393,22 +1393,12 @@ def test_transactions(conn: snowflake.connector.SnowflakeConnection):
         assert cur.fetchall() == [("Statement executed successfully.",)]
 
 
-def test_trim_cast_varchar(conn: snowflake.connector.SnowflakeConnection):
-    with conn.cursor() as cur:
-        cur.execute("create or replace table trim_cast_varchar(id number, name varchar);")
-        cur.execute("insert into trim_cast_varchar(id, name) values (1, '  name 1  '), (2, 'name2   ');")
-        cur.execute("select trim(id), trim(name) from trim_cast_varchar;")
+def test_trim_cast_varchar(cur: snowflake.connector.cursor.SnowflakeCursor):
+    cur.execute("select trim(1), trim('  name 1  ')")
+    assert cur.fetchall() == [("1", "name 1")]
 
-        assert cur.fetchall() == [("1", "name 1"), ("2", "name2")]
-
-    with conn.cursor() as cur:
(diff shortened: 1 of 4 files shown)


tests/test_fakes.py
```diff
@@ -1255,6 +1255,24 @@ def test_transactions(conn: snowflake.connector.SnowflakeConnection):
...
+        cur.execute("insert into trim_cast_varchar(id, name) values (1, '  name 1  '), (2, 'name2   ');")
+        cur.execute("select trim(id), trim(name) from trim_cast_varchar;")
+
+        assert cur.fetchall() == [("1", "name 1"), ("2", "name2")]
+
+    with conn.cursor() as cur:
+        cur.execute("create or replace table trim_cast_varchar_variant_field(data variant);")
+        cur.execute(
+            """insert into trim_cast_varchar_variant_field(data) values ('{"k1": "   v11  "}'),('{"k1": 21}');"""
+        )
```
tekumara, line 1270: This fails for me on a real Snowflake instance with:
Expression type does not match column data type, expecting VARIANT but got VARCHAR(18) for column DATA

Am guessing we need to convert it to variant first?

seruman: My bad again, updated.

As on #78 tests seems to fail due to -guessing- sqlalchemy version
https://github.com/tekumara/fakesnow/pull/78#discussion_r1566809604

PR #72, opened 2024-03-28
feat: cast string literals to timestamp in dateadd

seruman:
Snowflake -again- seems to cast string literal in `DATEADD`  to `TIMESTAMP_LTZ` or `TIMESTAMP_NTZ` depending on `TIMESTAMP_TYPE_MAPPING` session parameter.

```sql
select
  dateadd('day', 3, '2023-04-02') as c1, SYSTEM$TYPEOF(c1),
  dateadd('week', 3, '2023-04-02') as c3, SYSTEM$TYPEOF(c3),
  dateadd('month', 3, '2023-04-02') as c5, SYSTEM$TYPEOF(c5),
  dateadd('year', 3, '2023-04-02') as c7, SYSTEM$TYPEOF(c7),
;
```

|              C1               |   SYSTEM$TYPEOF(C1)    |
|-------------------------------|------------------------|
| 2023-04-05 00:00:00.000 +0000 | TIMESTAMP_LTZ(9)[SB16] |

|              C3               |   SYSTEM$TYPEOF(C3)    |
|-------------------------------|------------------------|
| 2023-04-23 00:00:00.000 +0000 | TIMESTAMP_LTZ(9)[SB16] |

|              C5               |   SYSTEM$TYPEOF(C5)    |
|-------------------------------|------------------------|
| 2023-07-02 00:00:00.000 +0000 | TIMESTAMP_LTZ(9)[SB16] |

|              C7               |   SYSTEM$TYPEOF(C7)    |
|-------------------------------|------------------------|
| 2026-04-02 00:00:00.000 +0000 | TIMESTAMP_LTZ(9)[SB16] |

Here seems to me we can cast the operand to `TIMESTAMP` if it is a string literal to achieve similar behaviour.

But I'm a little bit concerned on `TIMESTAMP` as DuckDB is quite picky operations between `TIMESTAMP` and `TIMESTAMP WITH TIMEZONE`. Like; `CURRENT_TIMESTAMP` returns `TIMESTAMP WITH TIMEZONE` and `DATE_DIFF` expects either operands to be same typed. I saw that `to_timestamp` transformation casts to `TIMESTAMP` and went with it.

### DIFF
--- a/fakesnow/transforms.py
+++ b/fakesnow/transforms.py
@@ -184,7 +184,7 @@ def dateadd_string_literal_timestamp_cast(expression: exp.Expression) -> exp.Exp
         "this",
         exp.Cast(
             this=expression.this,
-            # TODO(selman): TIMESTAMP_TZ OR NTZ?
+            # TODO: support TIMESTAMP_TYPE_MAPPING of TIMESTAMP_LTZ/TZ
             to=exp.DataType(this=exp.DataType.Type.TIMESTAMP, nested=False, prefix=False),
         ),
     )
--- a/fakesnow/transforms.py
+++ b/fakesnow/transforms.py
@@ -169,6 +169,33 @@ def drop_schema_cascade(expression: exp.Expression) -> exp.Expression:
     return new
 
 
+def dateadd_string_literal_timestamp_cast(expression: exp.Expression) -> exp.Expression:
+    """Snowflake's DATEADD function implicitly casts string literals to
+    timestamps regardless of unit.
+    """
+    if not isinstance(expression, exp.DateAdd):
+        return expression
+
+    dateadd = expression
+
+    if not isinstance(dateadd.this, exp.Literal) or not dateadd.this.is_string:
+        return expression
+
+    string_literal = dateadd.this
+
+    new_dateadd = dateadd.copy()
+    new_dateadd.set(
+        "this",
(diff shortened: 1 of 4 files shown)


fakesnow/transforms.py
```diff
@@ -169,6 +169,29 @@ def drop_schema_cascade(expression: exp.Expression) -> exp.Expression:
...
 
 
+def dateadd_string_literal_timestamp_cast(expression: exp.Expression) -> exp.Expression:
+    """Snowflake's DATEADD function implicitly casts string literals to
+    timestamps regardless of unit.
+    """
+    if not isinstance(expression, exp.DateAdd):
+        return expression
+
+    if not isinstance(expression.this, exp.Literal) or not expression.this.is_string:
```
tekumara, line 179: NB: It's a big future feature, but at some point if we had all the column lineage exposed, we could support column expressions that were VARCHAR too.

tekumara (commented):
Thank you!

fakesnow/transforms.py
seruman: that might be helpful on timestamp casting as well. snowflake's `cast( ... timestamp)` and `to_timestamp` both support conversion from numeric and varchar values, but on duckdb `cast` is solely for varchar and `to_timestamp` for numeric/bigint.

PR #73, opened 2024-03-29
feat: add timeadd as an alias for dateadd

seruman:
`sqlglot~=21.2.0` does not support `timeadd` for Snowflake, this transform simply tries to convert it to `DateAdd`.

[sqlglot v23.1.0](https://github.com/tobymao/sqlglot/blob/main/CHANGELOG.md#v2310---2024-03-26) adds `timeadd/addtime` with https://github.com/tobymao/sqlglot/pull/3180/files but upgrade would require changes on other transforms like `to_timestamp` and ones with json extraction.

### DIFF
--- a/fakesnow/fakes.py
+++ b/fakesnow/fakes.py
@@ -182,6 +182,7 @@ class FakeSnowflakeCursor:
             .transform(transforms.regex_replace)
             .transform(transforms.regex_substr)
             .transform(transforms.values_columns)
+            .transform(transforms.timeadd)
             .transform(transforms.to_date)
             .transform(transforms.to_decimal)
             .transform(transforms.to_timestamp_ntz)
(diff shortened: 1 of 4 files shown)


tekumara:
Ah it's probably best to do the upgrade of sqlglot. I try to keep it on the latest but have fallen behind.

seruman:
> Ah it's probably best to do the upgrade of sqlglot. I try to keep it on the latest but have fallen behind.

Might be better, on my fork I've bumped it to v22.4.0 for following changes;
https://github.com/tobymao/sqlglot/commit/c51b64fa6a437698fd8b347d98ffaf9fb543d2d5
https://github.com/tobymao/sqlglot/commit/ddab9dff663985d9473ce4b2dbe4fe266ae1bdf7
https://github.com/tobymao/sqlglot/commit/c333017fe49c0645cdaa3a75d0a7cc6a5b46dddc


I can close this PR and work on the upgrade. sqlglot is evolving in a fast pace, do you have a preferred version or the latest is ok?

tekumara:
That would be great, thankyou! The latest and greatest version 🙂

PR #74, opened 2024-03-29
fix: remove only null key/values in object_construct

seruman:
sqlglot transforms Snowlake's `zeroifnull` to `case when .. is null ...` statement. In `object_construct`, removing every `NULL` expression generates an unexpected query;

```sql
-- snowflake
SELECT OBJECT_CONSTRUCT( 'k1', 'v1', 'k2', CASE WHEN ZEROIFNULL(col) + 1 >= 2 THEN 'v2' ELSE NULL END, 'k3', 'v3')

-- before
SELECT TO_JSON({'k1': 'v1', 'k2': , 'k3': 'v3'})

-- after
SELECT TO_JSON({'k1': 'v1', 'k2': CASE WHEN CASE WHEN col IS NULL THEN 0 ELSE col END + 1 >= 2 THEN 'v2' ELSE NULL END, 'k3': 'v3'})
--                                          |------zeroifnull(col)-----|

```

AFAIS `Struct` only contains `EQ` expressions on `object_construct`. Instead of removing every `NULL` we can just remove `NULL` keyed/valued `EQ` expressions.

### DIFF
--- a/fakesnow/transforms.py
+++ b/fakesnow/transforms.py
@@ -541,12 +541,26 @@ def object_construct(expression: exp.Expression) -> exp.Expression:
     """
 
     if isinstance(expression, exp.Struct):
-        # remove expressions containing NULL
-        for enull in expression.find_all(exp.Null):
-            if enull.parent:
-                enull.parent.pop()
+        non_null_expressions = []
+        for e in expression.expressions:
+            if not (isinstance(e, exp.EQ)):
+                non_null_expressions.append(e)
+                continue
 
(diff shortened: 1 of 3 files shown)


seruman:
`EQ` nodes in `Struct` is replaced by `PropertyEQ` here; https://github.com/tobymao/sqlglot/commit/238f9aa7c32037d0c280cfe6ece77eed9c311cc5

requires changes after sqlglot upgrade.

tekumara:
Thanks!

tekumara:
Thanks, looks great!

PR #77, opened 2024-04-02
fix(json extraction): only return string when casting to varchar

seruman:
Not sure if intentional but despite doc comment `Return raw unquoted string when casting json extraction to varchar`, `json_extract_cast_as_varchar` transforms all CASTs on JSON extraction to `VARCHAR`.

```sql
-- snowflake
select parse_json('{"fruit":"9000"}'):fruit::number

-- becomes in duckdb;
SELECT JSON('{"fruit":"9000"}') ->> '$.fruit'
```

### DIFF
--- a/tests/test_fakes.py
+++ b/tests/test_fakes.py
@@ -1450,6 +1450,17 @@ def test_values(conn: snowflake.connector.SnowflakeConnection):
         ]
 
 
+def test_json_extract_cast_as_varchar(dcur: snowflake.connector.cursor.DictCursor):
+    dcur.execute("CREATE TABLE example (j VARIANT)")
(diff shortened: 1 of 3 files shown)


tests/test_fakes.py
```diff
@@ -1312,6 +1312,17 @@ def test_values(conn: snowflake.connector.SnowflakeConnection):
         ]
 
 
+def test_json_extract_cast_as_varchar(dcur: snowflake.connector.cursor.DictCursor):
+    dcur.execute("CREATE TABLE example (j VARIANT)")
+    dcur.execute("""INSERT INTO example VALUES ('{"str": "100", "number" : 100}')""")
```
tekumara, line 1317: I get this error on a real Snowflake instance:
Expression type does not match column data type, expecting VARIANT but got VARCHAR(30) for column J

PR #78, opened 2024-04-02
fix: GET_PATH precedence for JSONExtractScalar

seruman:
I've encountered another precedence issue with json extract scalar this time;

```sql
select JSON('{"K1": "a", "K2": "b"}') as col, CASE WHEN COL ->> '$.K1' = 'a' AND COL ->> '$.K2' = 'b' THEN 'yes' END as r;
-- Error: Conversion Error: Failed to cast value to numerical: {"K1":"a","K2":"b"}

select JSON('{"K1": "a", "K2": "b"}') as col, CASE WHEN (COL ->> '$.K1') = 'a' AND (COL ->> '$.K2') = 'b' THEN 'yes' END as r;
-- ┌─────────────────────┬─────────┐
-- │         col         │    r    │
-- │        json         │ varchar │
-- ├─────────────────────┼─────────┤
-- │ {"K1":"a","K2":"b"} │ yes     │
-- └─────────────────────┴─────────┘

```

### DIFF
--- a/tests/test_fakes.py
+++ b/tests/test_fakes.py
@@ -797,7 +797,7 @@ def test_get_path_precedence(cur: snowflake.connector.cursor.SnowflakeCursor):
 
     cur.execute("create table example (col variant)")
 
-    cur.execute("""insert into example values ('{"K1": "a", "K2": "b"}')""")
+    cur.execute("""insert into example select parse_json('{"K1": "a", "K2": "b"}')""")
     cur.execute("select case when col:K1::VARCHAR = 'a' and col:K2::VARCHAR = 'b' then 'yes' end from example")
     assert cur.fetchall() == [("yes",)]
 
--- a/tests/test_fakes.py
+++ b/tests/test_fakes.py
@@ -795,11 +795,10 @@ def test_get_path_precedence(cur: snowflake.connector.cursor.SnowflakeCursor):
     cur.execute("select {'K1': {'K2': 1}} as col where col:K1:K2 > 0")
     assert indent(cur.fetchall()) == [('{\n  "K1": {\n    "K2": 1\n  }\n}',)]
 
-    cur.execute("create table example (col variant)")
-
-    cur.execute("""insert into example select parse_json('{"K1": "a", "K2": "b"}')""")
-    cur.execute("select case when col:K1::VARCHAR = 'a' and col:K2::VARCHAR = 'b' then 'yes' end from example")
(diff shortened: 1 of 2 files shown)


tests/test_fakes.py
```diff
@@ -707,6 +707,15 @@ def test_get_path_precedence(cur: snowflake.connector.cursor.SnowflakeCursor):
     cur.execute("select {'K1': {'K2': 1}} as col where col:K1:K2 > 0")
     assert indent(cur.fetchall()) == [('{\n  "K1": {\n    "K2": 1\n  }\n}',)]
 
+    cur.execute("select {'K1': {'K2': 1}} as col where col:K1:K2 > 0")
+    assert indent(cur.fetchall()) == [('{\n  "K1": {\n    "K2": 1\n  }\n}',)]
+
+    cur.execute("create table example (col variant)")
+
+    cur.execute("""insert into example values ('{"K1": "a", "K2": "b"}')""")
```
tekumara, line 715: This gives me the following on the real Snowflake:
Expression type does not match column data type, expecting VARIANT but got VARCHAR(22) for column COL

seruman: Weirdly I've made mistakes moving queries from Snowflake worksheet to tests. 

What is your strategy on testing against real Snowflake?

seruman: Btw, tests fails due to sqlalchemty, guessing something to do with `snowflake-sqlalchemy` ?

seruman: Not sure which versions get installed on cached venv previously, but on my local machine it was;
```
snowflake-sqlalchemy==1.5.1
SQLAlchemy==1.4.52
```

on failing run;

```
snowflake-sqlalchemy==1.5.2
SQLAlchemy==2.0.29
```

tekumara: > What is your strategy on testing against real Snowflake?

I copy-paste the test into [snowflake.ipynb](https://github.com/tekumara/fakesnow/blob/main/notebooks/snowflake.ipynb) and run it there.

tekumara: Looks like snowflake-sqlalchemy unpinned SQLAlchemy but its now fixed in [1.5.3](https://github.com/snowflakedb/snowflake-sqlalchemy/releases/tag/v1.5.3). So I reran the GHA workflow and its now passed.

tekumara:
Thank you for the fix!

seruman:
> Thank you for the fix!

thanks for this awesome project 🎉

PR #79, opened 2024-04-02
feat: cast string literals to timestamp in datediff

seruman:
Basically the same as https://github.com/tekumara/fakesnow/pull/72 but for `DATEDIFF`. 

Same concerns of the `DATEADD` change applies here;
> But I'm a little bit concerned on TIMESTAMP as DuckDB is quite picky operations between TIMESTAMP and TIMESTAMP WITH TIMEZONE. Like; CURRENT_TIMESTAMP returns TIMESTAMP WITH TIMEZONE and DATE_DIFF expects either operands to be same typed. I saw that to_timestamp transformation casts to TIMESTAMP and went with it.


tekumara:
👍

PR #96, opened 2024-05-31
feat: Support timestamp_ntz.

DanCardin:
Do you have a suggestion/preference on how I should go about adding support for this? I have the failing test, but although i see "timestamp_ntz" referenced in the repo, not so whatever would be required to support it at the table creation level.

### DIFF
--- a/tests/test_sqlalchemy.py
+++ b/tests/test_sqlalchemy.py
@@ -1,5 +1,7 @@
+import datetime
 from typing import cast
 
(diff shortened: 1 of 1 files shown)


tekumara:
Hi @DanCardin I've added support for TIMESTAMP_NTZ in 0493ce6

DanCardin:
Ah, can confirm that release does fix my issue. thanks!

PR #100, opened 2024-06-04
fix: ALTER TABLE result output

DanCardin:
fwiw, I'm attempting to use fakesnow to test alembic migrations of snowflake 😬.

You already have a test for `alter table`, but sqlalchemy (versus the raw cursor) always calls `.description` on the results of queries and it was hitting similar issues to issues you've already solved with e.g. `CREATE TABLE` etc.

I deviated from the pattern a little bit because `AlterTable` constructs returned `ALTERTABLE` as the `cmd`. It seemed somewhat unfortunate to compare exactly against that, but also seemed very likely that other ALTER statements would have the same generic output.

### DIFF
--- a/fakesnow/fakes.py
+++ b/fakesnow/fakes.py
@@ -297,6 +297,9 @@ class FakeSnowflakeCursor:
             elif cmd == "CREATE TABLE" and ident:
                 result_sql = SQL_CREATED_TABLE.substitute(name=ident)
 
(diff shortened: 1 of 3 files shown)


tekumara:
Thanks for this. I've moved the test into `test_description_create_alter_drop_table` with the related tests, rather than make it as sqlalchemy specific thing.

PR #103, opened 2024-06-21
fix: CREATE DATABASE when using db_path

jsibbison-square:
When patching in a test specifying a db_path and attempting to create a database:
```
with fakesnow.patch(db_path="databases"):
    cursor = snowflake.connector.connect().cursor()
    cursor.execute(f"CREATE DATABASE TEST32cs3s3")
```
the following error is thrown
```
>           db_file = f"{db_path/db_name}.db" if db_path else ":memory:"
E           TypeError: unsupported operand type(s) for /: 'str' and 'str'
```

This PR fixes the string interpolation.

Note that I expect there is a similar bug in [fakes.py#538](https://github.com/jazzarati/fakesnow/blob/14e6a41bdc133ebc86630467dcaf997c943dfd8a/fakesnow/fakes.py#L538) but I couldn't figure out how to trigger it via a test.

PS: Thanks for making this project

### DIFF
--- a/fakesnow/fakes.py
+++ b/fakesnow/fakes.py
@@ -520,7 +520,7 @@ class FakeSnowflakeConnection:
         self.schema = schema and schema.upper()
         self.database_set = False
         self.schema_set = False
-        self.db_path = db_path
+        self.db_path = Path(db_path) if db_path else None
         self.nop_regexes = nop_regexes
         self._paramstyle = snowflake.connector.paramstyle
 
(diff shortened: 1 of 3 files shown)


tekumara:
Good catch thank you! The `/` operator will work with `Path` but not `str` which is why [fakes.py#538](https://github.com/jazzarati/fakesnow/blob/14e6a41bdc133ebc86630467dcaf997c943dfd8a/fakesnow/fakes.py#L538) works.

I've refactored your solution to use `Path` as this is the type expected by [transforms.create_database](https://github.com/jazzarati/fakesnow/blob/14e6a41bdc133ebc86630467dcaf997c943dfd8a/fakesnow/transforms.py#L99).

PR #108, opened 2024-06-26
fix: Support IF NOT EXISTS in CREATE DATABASE statements

jsibbison-square:
If a database already existed and we try to execute "CREATE DATABASE IF NOT EXISTS DB2" then it currently fails with error:
```
snowflake.connector.errors.ProgrammingError: 002043 (02000): Binder Error: Unique file handle conflict: Database "DB2" is already attached with path "...fakesnow-testl1jfag6s/DB2.db"
```

This commit adds support for this functionality by converting the statement to the [duckdb syntax of ATTACH IF NOT EXISTS](https://duckdb.org/docs/sql/statements/attach) when the exists clause is present.

Up to you if you want to keep the test in `test_fakes.py` or only use the one in `test_transforms.py`

### DIFF
--- a/fakesnow/transforms.py
+++ b/fakesnow/transforms.py
@@ -119,9 +119,13 @@ def create_database(expression: exp.Expression, db_path: Path | None = None) -> 
         db_name = ident.this
         db_file = f"{db_path/db_name}.db" if db_path else ":memory:"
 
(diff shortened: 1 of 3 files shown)


tekumara:
Thank you!

PR #111, opened 2024-06-30
feat: Implements basic snowflake session variables via SET/UNSET

jsibbison-square:
Implements the very basic usages of sql variables
```
SET var2 = 'blue';
SELECT * from table where color = $var2;                  Produces: SELECT * from table where color = 'blue';
UNSET var2;
```
as described in https://docs.snowflake.com/en/sql-reference/session-variables.

List of features supported and not supported by this PR:

- [x] Variables are scoped to the session (Eg. The connection, not the cursor)
- [x] Simple scalar variables: SET var1 = 1;
- [x] Unset variables: UNSET var1;
- [x] Simple SQL expression variables: SET INCREMENTAL_DATE = DATEADD( 'DAY', -7, CURRENT_DATE());
- [x] Basic use of variables in SQL using $ syntax: SELECT $var1;
- [ ] Multiple variables: SET (var1, var2) = (1, 'hello');
- [ ] Variables set via 'properties' on the connection https://docs.snowflake.com/en/sql-reference/session-variables#setting-variables-on-connection
- [ ] Using variables via the IDENTIFIER function: INSERT INTO IDENTIFIER($my_table_name) (i) VALUES (42);
- [ ] Session variable functions: https://docs.snowflake.com/en/sql-reference/session-variables#session-variable-functions

### DIFF
--- a/fakesnow/fakes.py
+++ b/fakesnow/fakes.py
@@ -33,6 +33,7 @@ import fakesnow.info_schema as info_schema
 import fakesnow.macros as macros
 import fakesnow.transforms as transforms
 from fakesnow.global_database import create_global_database
+from fakesnow.variables import Variables
 
 SCHEMA_UNSET = "schema_unset"
 SQL_SUCCESS = "SELECT 'Statement executed successfully.' as 'status'"
@@ -141,9 +142,6 @@ class FakeSnowflakeCursor:
             else:
                 expression = parse_one(command, read="snowflake")
                 transformed = self._transform(expression)
-                if Variables.is_variable_modifier(transformed):
-                    self._conn.variables.update_variables(transformed)
-                    transformed = transforms.SUCCESS_NOP  # Nothing further to do if its a SET/UNSET operation.
             return self._execute(transformed, params)
         except snowflake.connector.errors.ProgrammingError as e:
             self._sqlstate = e.sqlstate
@@ -152,6 +150,7 @@ class FakeSnowflakeCursor:
     def _transform(self, expression: exp.Expression) -> exp.Expression:
         return (
             expression.transform(transforms.upper_case_unquoted_identifiers)
+            .transform(transforms.update_variables, variables=self._conn.variables)
             .transform(transforms.set_schema, current_database=self._conn.database)
             .transform(transforms.create_database, db_path=self._conn.db_path)
             .transform(transforms.extract_comment_on_table)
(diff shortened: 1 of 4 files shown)


tests/test_fakes.py
```diff
@@ -1542,6 +1543,54 @@ def test_use_invalid_schema(_fakesnow: None):
...
+        cur.execute("INSERT INTO example VALUES (10, 'hello'), (20, 'world');")
+        cur.execute("select id, name from example where name = $var2;")
+        assert cur.fetchall() == [(10, "hello")]
+
+        cur.execute("UNSET var3;")
+        try:
+            cur.execute("select $var3;")
+            raise AssertionError("Expected exception")
+        except duckdb.duckdb.InvalidInputException:
+            pass
```
tekumara, line 1578: To test for an exception I'd use [pytest.raises](https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions), eg:
```suggestion
        with pytest.raises(ProgrammingError, match='Session variable '$VAR3' does not exist'):
        cur.execute("select $var3;")
```

Note also fakesnow should raise the snowflake `ProgrammingError` rather than a duckdb error, to avoid the duckdb implementation leaking out into clients.

And optionally have a similar error message to Snowflake eg: 
> Session variable '$VAR3' does not exist

tests/test_fakes.py
```diff
@@ -1542,6 +1543,54 @@ def test_use_invalid_schema(_fakesnow: None):
...
+
+    # variables are scoped to the session, so they should be available in a new cursor.
+    with conn.cursor() as cur:
+        cur.execute("select $var1, $var2, $var4;")
+        assert cur.fetchall() == [(1, "hello", [1, 2, 3])]
```
tekumara, line 1583:
```suggestion
        assert cur.fetchall() == [(1, "hello", '[1,2,3]')]
```

fakesnow/fakes.py
```diff
@@ -134,12 +134,16 @@ def execute(
...
+            command = self._inline_variables(command)
             command, params = self._rewrite_with_params(command, params)
             if self._conn.nop_regexes and any(re.match(p, command, re.IGNORECASE) for p in self._conn.nop_regexes):
                 transformed = transforms.SUCCESS_NOP
             else:
                 expression = parse_one(command, read="snowflake")
                 transformed = self._transform(expression)
+                if Variables.is_variable_modifier(transformed):
+                    self._conn.variables.update_variables(transformed)
+                    transformed = transforms.SUCCESS_NOP  # Nothing further to do if its a SET/UNSET operation.
```
tekumara, line 146: (optional) for consistency I think this could be handled like any other transform, eg: in `transforms.py` as the method
```python
def update_variables(
    expression: exp.Expression,
    variables: dict[str, Any],
) -> exp.Expression:
    # if is variable modifier, update variables dict and return transforms.SUCCESS_NOP
```

tekumara (commented):
This is a great feature, thanks for implementing this! Some suggestions (including optional ones) below

jsibbison-square:
These comments make sense. I'll work through these changes and notify you when its ready for review again.

jsibbison-square:
@tekumara I've made updates from all the comments I believe this is ready for review again :).

tests/test_fakes.py
```diff
@@ -1542,6 +1543,54 @@ def test_use_invalid_schema(_fakesnow: None):
...
+# [ ] Multiple variables: SET (var1, var2) = (1, 'hello');
+# [ ] Variables set via 'properties' on the connection https://docs.snowflake.com/en/sql-reference/session-variables#setting-variables-on-connection
+# [ ] Using variables via the IDENTIFIER function: INSERT INTO IDENTIFIER($my_table_name) (i) VALUES (42);
+# [ ] Session variable functions: https://docs.snowflake.com/en/sql-reference/session-variables#session-variable-functions
+def test_variables(conn: snowflake.connector.SnowflakeConnection):
+    with conn.cursor() as cur:
+        cur.execute("SET var1 = 1;")
+        cur.execute("SET var2 = 'hello';")
+        cur.execute('SET var3 = parse_json(\'{"k1": "v1"}\');')
+        cur.execute("SET var4 = array_construct(1, 2, 3);")
```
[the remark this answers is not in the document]
jsibbison-square: Can't believe I missed this initially 🤦 .

tekumara:
Prefect thank you!

PR #114, opened 2024-07-10
feat: SHOW PRIMARY KEYS for table

Guillem96:
Previously there was a `NotImplemented` error when running a query to request the PKs of a table.

I added support for listing the PKs of a table.


tekumara:
Perfect thank you!

PR #116, opened 2024-07-11
fix: Only set variables for SetItem expressions

jsibbison-square:
The existing logic to set a variable was being triggered by some other sql that includes the SET keyword.

I found this case when working on MERGE which has the statement WHEN MATCHED THEN SET ... This triggered the issue. 

Unfortunately we don't currently support MERGE and UPDATE TABLE SET ... statements don't seem to trigger this issue so I couldn't find an existing supported usecase to add a test so there are no new tests.

Once MERGE is built it will fail without this change so its a future test case that will cover us.

In this PR I also inverted the expression and fixed some bad variable naming that was confusing.


tekumara (approved):
👍

PR #121, opened 2024-08-02
fix: Ensure type column in information_schema views is not null

DanCardin:
I noticed this when trying to use snowflake-sqlalchemy to test view creation using fakesnow. It's actually sort of also a bug in snowflake-sqlalchemy, that it's hitting an untested code path and failing. but also, this column **should** have a type, and adding one avoids the latent bug in snowflake-sqlalchemy.

### DIFF
--- a/tests/test_info_schema.py
+++ b/tests/test_info_schema.py
@@ -217,7 +217,7 @@ def test_type_column_is_not_null(cur: snowflake.connector.cursor.SnowflakeCursor
     for table in [
         "information_schema.databases",
         "information_schema.views",
-        "information_schema._fs_columns_snowflake",
+        "information_schema.columns",
     ]:
         cur.execute(f"DESCRIBE {table}")
         result = cur.fetchall()
--- a/tests/test_info_schema.py
+++ b/tests/test_info_schema.py
@@ -213,7 +213,7 @@ def test_info_schema_show_primary_keys_from_table(cur: snowflake.connector.curso
     assert pk_columns == ["ID", "VERSION"]
 
 
-def test_null_data_type(cur: snowflake.connector.cursor.SnowflakeCursor) -> None:
+def test_type_column_is_not_null(cur: snowflake.connector.cursor.SnowflakeCursor) -> None:
     for table in [
         "information_schema.databases",
         "information_schema.views",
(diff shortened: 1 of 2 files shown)


tests/test_info_schema.py
```diff
@@ -211,3 +211,16 @@ def test_info_schema_show_primary_keys_from_table(cur: snowflake.connector.curso
...
     assert pk_columns == ["ID", "VERSION"]
+
+
+def test_null_data_type(cur: snowflake.connector.cursor.SnowflakeCursor) -> None:
+    for table in [
+        "information_schema.databases",
+        "information_schema.views",
+        "information_schema._fs_columns_snowflake",
+    ]:
+        cur.execute(f"DESCRIBE {table}")
```
tekumara, line 222: Needed for a real snowflake instance:
```suggestion
   cur.execute(f"DESCRIBE VIEW {table}")
```

tests/test_info_schema.py
```diff
@@ -211,3 +211,16 @@ def test_info_schema_show_primary_keys_from_table(cur: snowflake.connector.curso
 
     pk_columns = [result[4] for result in pk_result]
     assert pk_columns == ["ID", "VERSION"]
+
+
+def test_null_data_type(cur: snowflake.connector.cursor.SnowflakeCursor) -> None:
+    for table in [
+        "information_schema.databases",
+        "information_schema.views",
+        "information_schema._fs_columns_snowflake",
```
tekumara, line 220: A real snowflake instance has:
```suggestion
        "information_schema.columns",
```

tekumara (commented):
Good catch!

tests/test_info_schema.py
```diff
@@ -211,3 +211,16 @@ def test_info_schema_show_primary_keys_from_table(cur: snowflake.connector.curso
 
     pk_columns = [result[4] for result in pk_result]
     assert pk_columns == ["ID", "VERSION"]
+
+
+def test_null_data_type(cur: snowflake.connector.cursor.SnowflakeCursor) -> None:
```
tekumara, line 216:
```suggestion
def test_type_column_is_not_null(cur: snowflake.connector.cursor.SnowflakeCursor) -> None:
```

PR #122, opened 2024-08-03
fix: Allow connection with schema=information_schema

DanCardin:
With snowflake, it's sometimes (particularly with alembic) convenient to connect with `schema='information_schema'`, so that all other schemas are required explicitly (since there's no public schema, and the weird way snowflake scoping works).

As-is without this PR, the duckdb `information_schema` is lowercased, and the auto-use-schema code in fakesnow would unconditionally `.upper()` the given schema, search for `INFORMATION_SCHEMA`, not find it, and then you'd get `This session does not have a current schema. Call 'USE SCHEMA'` when you wouldn't in real life.

It might perhaps be more ideal to somehow fully recreate the snowflake information_schema and replace it, because there **are** some differences with the duckdb version that i suspect will end up being irreconcilable, but it's not obvious to me **how** to do this, nor how much additional work it'd be so...just a thought.

### DIFF
--- a/tests/test_connect.py
+++ b/tests/test_connect.py
@@ -83,6 +83,15 @@ def test_connect_db_path_reuse():
             assert cur.execute("select * from example").fetchall() == [(420,)]
 
 
+def test_connect_information_schema():
+    with fakesnow.patch(create_schema_on_connect=False):
+        conn = snowflake.connector.connect(database="db1", schema="information_schema")
+        assert conn.schema == "INFORMATION_SCHEMA"
+        with conn, conn.cursor() as cur:
+            # shouldn't fail
+            cur.execute("SELECT * FROM databases")
+
+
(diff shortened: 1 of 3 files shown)


tests/test_connect.py
```diff
@@ -83,6 +83,15 @@ def test_connect_db_path_reuse():
             assert cur.execute("select * from example").fetchall() == [(420,)]
 
 
+def test_connect_information_schema():
+    with fakesnow.patch(create_schema_on_connect=False):
+        conn = snowflake.connector.connect(database="db1", schema="information_schema")
+        assert conn.schema == "INFORMATION_SCHEMA"
```
tekumara, line 89: Thank you! I've improved the implementation to support this check as per a real Snowflake instance

tekumara (approved):

tekumara:
PS: yes having a fully recreated snowflake info schema makes sense to improve fidelity.

PR #144, opened 2024-10-30
feat: Support TRUNCATE TABLE description

DanCardin:
I did this differently from all of the other describe operations because `describe truncate` is (currently) a syntax error in sqlglot, and it didn't seem to be of particular import to fakesnow whether it **was** supported or not.

In fact, (to me) it seems like dealing with the `describe` special cases in this location, in general, will result in less special case code handing for the majority case of non-describe sql statements, which seems like a win to me?

### DIFF
--- a/tests/test_fakes.py
+++ b/tests/test_fakes.py
@@ -1522,3 +1522,24 @@ def test_json_extract_cast_as_varchar(dcur: snowflake.connector.cursor.DictCurso
 
     dcur.execute("SELECT j:str::number as j_str_number, j:num::number as j_num_number FROM example")
     assert dcur.fetchall() == [{"J_STR_NUMBER": 100, "J_NUM_NUMBER": 200}]
+
+
+def test_truncate(dcur: snowflake.connector.cursor.DictCursor):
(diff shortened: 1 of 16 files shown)


tests/test_fakes.py
```diff
@@ -1522,3 +1522,24 @@ def test_json_extract_cast_as_varchar(dcur: snowflake.connector.cursor.DictCurso
...
+        ResultMetadata(
+            name="status",
+            type_code=2,
+            display_size=None,
+            internal_size=16777216,
+            precision=None,
+            scale=None,
+            is_nullable=True,
+        )
+    ]
```
tekumara, line 1542: To match the snowflake result set this would need to assert:
```suggestion
    assert dcur.fetchall() == [{"status": "Statement executed successfully."}]
```

To do this, try adding another branch [here](https://github.com/dancardin/fakesnow/blob/9ecb6dca5b632c3feced47cc65e8c5e070a3e56d/fakesnow/cursor.py#L318) which sets `result_sql = SQL_SUCCESS`.

You won't need to change `_describe_last_sql`

Issue #35, opened 2024-01-23
Support for loading data from stages


tharwan:
Hi,

Would it be possible to support loading of data from a stage? (https://docs.snowflake.com/en/user-guide/data-load-considerations-load)

I am not sure from the docs if this would be even inside the scope of this project. 

However it would help us a lot to test our complete ETL workflow.

tekumara:
Hi @tharwan this might be possible ... could you share the SQL statements for both creating up the stage and loading from it you'd like support for?

tekumara:
Hi @tharwan this might be possible ... could you share the SQL statements you'd like support for, both creating the stage and loading from it?

tekumara:
Hi @tharwan this might be possible ... could you share example SQL statements you'd like support for, both creating the stage and loading from it?

tharwan:
Loading from a stage looks something like this:

```SQL
COPY INTO my_table
FROM @my_stage/my_file
FILE_FORMAT = (FIELD_DELIMITER = ';')
```

Creating a stage is not so relevant in our case but looks like this:
```SQL
CREATE STAGE my_ext_stage
URL='azure://myaccount.blob.core.windows.net/load/files/'
STORAGE_INTEGRATION = myint;
```

tekumara:
Thanks! From the above I gather you want support for CSV files.

I think creating the stage would be needed for fakesnow to know where to find the files to load.

tharwan:
another option could be to just look locally for a file and ignore the @my_stage part. 

e.g. 

```SQL
COPY INTO my_table
FROM @my_stage/my_file
FILE_FORMAT = (FIELD_DELIMITER = ';')
```
would translate to

```SQL
COPY my_table
FROM my_file
(FORMAT CSV, DELIMITER ';');
```

DanCardin:
we recently found ourselves wanting this at $job. i may look into a relatively simple initial implementation somewhat soon. My personal plan/preference would be for fakesnow to simply internally track creations of stages inside `information_schema.stages` and something else for integrations (i dont see a table for this, but we're using integrations personally).

And then translate COPY statements against the service in question into boto3 calls + insert statements (which ought to work for all the supported cloud providers iirc)

It'd then be the job of the user to set up `moto` however they prefer, to avoid making calls to the actual service. Perhaps fakesnow could provide a default fixture that would ensure it'd avoid making calls by default, but i'd suspect we at least, would be forced to override it anyways.

---

[...]

PR #185, opened 2025-03-06
feat: basic COPY INTO .... FORCE = TRUE

DanCardin:
* Related to https://github.com/tekumara/fakesnow/issues/35, but wouldn't be all the way to fixing it, given that this does not handle stages yet
* ~~Refactors `transforms.py` into a package, given the size of the impl of this particular transform~~

   ~~Moves `transform_merge.py` into new page because it seems like it should live there. The existence of this module is what made me think a package was already a good idea. Plus the size of the existing module is 1500loc.~~
   
   ~~I'd be happy to break the root module up further, but wasn't even sure you'd like this refactor as-is. plus a larger unncessary diff for this feature.~~
* Implements a single, specific variant of the COPY INTO command, which ingests data from a S3 location rather than a stage reference.

  As you can see from the impl, a lot of options are either ignored or left unimplemented and fail, but this particular syntax is one of a few that we depend upon currently so i started here.

~~One issue right now is that this "cant" be properly tested afaict without running a moto server and configuring the duckdb connection to use a different S3 url. My personal inclination would be to use [pytest-mock-resources](https://github.com/schireson/pytest-mock-resources/) to instrument a moto docker container, just because i know it works and it's simple to set up.~~

But it's also a fairly opinionated change only necessary for this line of featuring, so figured I should wait to see your reaction to this first.


tests/test_transforms_copy_into.py
```diff
@@ -0,0 +1,30 @@
+import snowflake.connector.cursor
+import sqlglot
+
+from fakesnow.transforms.copy_into import copy_into
+
+example1 = """
+COPY INTO schema_.table_ (a, b)
+FROM "s3://bucket/"
```
tekumara, line 8: A real Snowflake instance will throw a SQL compilation error if double quotes are used
```suggestion
FROM 's3://bucket/'
```

tests/test_transforms_copy_into.py
```diff
@@ -0,0 +1,30 @@
...
+"""
+
+
+def test_transform_example1() -> None:
+    result = str(sqlglot.parse_one(example1, read="snowflake").transform(copy_into))
+
+    expected_result = (
+        "INSERT INTO schema_.table_ "
+        "SELECT a, b "
+        "FROM READ_CSV('s3://bucket/foo.txt', header = 1, quote = '\"', sep = ',')"
```
tekumara, line 22: `sep = ','` but `FIELD_DELIMITER` is `|` above - shouldn't these be the same?

tekumara (commented):
Thank you for this, great stuff! Overall looks good just two main things:

1. The refactoring is good, although I usually split these out into their own PRs. If there's more refactoring to do then let's do that in a separate PR.

2. Can we use the [moto server fixture](https://docs.getmoto.org/en/latest/docs/server_mode.html#start-within-python) rather than a docker container? It'll be more light weight and avoid making docker a requirement for local development and CI.

tests/test_transforms_copy_into.py
DanCardin: mm right. i haven't fully verified my current impl since i can't test the real instance yet.

DanCardin:
> Can we use the [moto server fixture](https://docs.getmoto.org/en/latest/docs/server_mode.html#start-within-python) rather than a docker container? It'll be more light weight and avoid making docker a requirement for local development and CI.

That's largely what I know the docker-based fixture is doing. e.g. spinning up a server and then configuring the fixture such that each test gets a unique, empty environment. Which are just things i wouldn't expect this library to really want to need to be thinking about.

I suppose I could look into an inline threaded mode for PMR such that it doesn't introduce an unnecessary reliance on docker for you.

tekumara:
Thanks @DanCardin happy to help out if you hit any snags.

DanCardin:
I inlined **a** solution derived from what's in pytest-mock-resources that uses the threaded server, just so it was obvious what was going on, and wasn't sure how much you'd want to rely on an unknown-to-you lib like that.

but i'd be equally happy adding support directly to it, and just consuming the fixtures here. lmk

tests/moto.py
```diff
@@ -0,0 +1,101 @@
...
+
+        credentials = response["Credentials"]
+
+        access_key_id = credentials["AccessKeyId"]
+        secret_access_key = credentials["SecretAccessKey"]
+        session_token = credentials["SessionToken"]
+        with conn.cursor() as cursor:
+            cursor.execute(
+                f"""
+                CREATE SECRET s3_secret (
```
tekumara, line 60: I was confused at first and thought this was valid snowflake sql lol!
Can we leave a comment that this is actually duckdb sql.

tests/moto.py
```diff
@@ -0,0 +1,101 @@
...
+                    TYPE s3,
+                    ENDPOINT '{moto_server.removeprefix("http://")}',
+                    KEY_ID '{access_key_id}',
+                    SECRET '{secret_access_key}',
+                    SESSION_TOKEN '{session_token}',
```
tekumara, line 65: Does it work without these? Might simplify things if it does.

tekumara (commented):
The moto server looks good - thankyou.

tests/moto.py
DanCardin: i dont think so. the point of this being how it is, is that otherwise by default all requests go to the same stateful moto account. whereas with this, every test gets a unique empty account that's guaranteed distinct from other tests.

there's presumably some way of resetting the state between tests within an account, but i dont like (personally) precluding the eventuality that you'd run your tests in parallel and have issues with tests clobbering one another.

tekumara: Ok, let's keep it. I have some ideas for how to hide the `Session` dataclass from the tests, and just expose an `s3_client` to them to simplify usage. And also make the moto fixtures more generic, and usable outside of this project, by moving the duckdb secret creation out. See fa21f30 which I pushed to your branch.

tests/test_transforms_copy_into.py
```diff
@@ -0,0 +1,90 @@
...
+    STORAGE_INTEGRATION = some_name
+    FILES=('foo.txt')
+    FILE_FORMAT = (TYPE = 'CSV')
+    FORCE = TRUE;
+    """
+
+    def test_transform(self) -> None:
+        result = str(sqlglot.parse_one(self.sql.format(bucket="bucket"), read="snowflake").transform(copy_into))
+
+        expected_result = "INSERT INTO schema_.table_ SELECT a, b FROM READ_CSV('s3://bucket/foo.txt', header = 1)"
```
tekumara, line 35: by default [snowflake doesn't skip the header](https://docs.snowflake.com/en/sql-reference/sql/copy-into-table), so we shouldn't either: 

CleanShot 2025-03-23 at 12 16 23@2x

I think this is why when I run this on a Snowflake instance I get:

```
RuntimeError: (snowflake.connector.errors.ProgrammingError) 100038 (22018): Numeric value 'a' is not recognized
  File 'delme.csv', line 1, character 1
  Row 1, column "TABLE1"["A":1]
  If you would like to continue loading when an error is encountered, use other values such as 'SKIP_FILE' or 'CONTINUE' for the ON_ERROR option. For more information on loading options, please run 'info loading_data' in a SQL client.
[SQL: COPY INTO table1 (a, b)
            FROM @delme
            FILE_FORMAT = (TYPE = 'CSV')
            FORCE = TRUE;]
```

DanCardin: if i'm interpreting where your point and the point in the code here correctly: `header=1` in duckdb enables header reading to determine the columns automatically, like snowflake (duckdb also has an option to skip rows like this though `skip=n`). I perhaps should have done `header=true`, but i was having issues with sqlglot's encoding of literals

tekumara: IIUC by default snowflake doesn't read the header at all, and so the first line is considered values, eg: in the error above it's trying to read the header `a` as a numeric into the table.

There's a `PARSE_HEADER = true` option in snowflake, which will use the header to determine column names. By default it's false. And it only applies when using the [INFER_SCHEMA function](https://docs.snowflake.com/en/sql-reference/functions/infer_schema). 

CleanShot 2025-03-25 at 10 37 10@2x

tekumara: I've implemented these semantics in [84a30fa](https://github.com/tekumara/fakesnow/pull/185/commits/84a30fa1a2c4fee50269aca4dfaca026a96f8201) plus some error handling and returning the load result set

tests/test_transforms_copy_into.py
tekumara: Updated in 84a30fa

tekumara:
Thank you @DanCardin for this, I've made some final changes in [84a30fa](https://github.com/tekumara/fakesnow/pull/185/commits/84a30fa1a2c4fee50269aca4dfaca026a96f8201) (header semantics, error handling and the load result set) to match Snowflake and I think it's good to go.

DanCardin:
Thanks for pushing this across the finish line! I was hoping to get back to it soon but this is better :P.

I likely **will** follow up with more options, since i'll be surprised if this covers all of our usecases.

PR #189, opened 2025-03-11
refactor: Move transforms into a package.

DanCardin:
Re https://github.com/tekumara/fakesnow/pull/185, you'd mentioned preferring such things to be in separate PRs. I'd already done that work as the first 2 commits of the PR.

and I figure this is going to be easier to merge more quickly, so it'll avoid unrelated conflicts on the COPY INTO feature, so it's in my own best interests to do it separately :P


tekumara:
Thank you :-)

PR #217, opened 2025-05-21
feat: Convert dollar characters in identifiers

denisrossetre:
* Transforms all dollar signs in Snowflake identifiers into underscores. Dollar signs are valid in [Snowflake identifiers](https://docs.snowflake.com/en/sql-reference/identifiers-syntax) but not in DuckDB.


denisrossetre:
Wrong approach. Identifiers containing a dollar sign are supported by DuckDB, the problem was that the variable interpolation was too eager. Fixed by another PR, closing this one.

PR #231, opened 2025-06-25
refactor: remove use of pandas

DanCardin:
Pandas does not appear to be a direct transitive dependency of fakesnow (and you seem to have an `if TYPE_CHECKING` block specifically to lazy-import pandas already).

So this top-level import seems to either be a missing required dependency, or it needs to be inlined/otherwise made lazy.

I'd have made it try/catch at the top level, but that messes up typing typically.

Afaict this is the only direct use of `pd` in the file outside type hints.

### DIFF
--- a/fakesnow/cursor.py
+++ b/fakesnow/cursor.py
@@ -10,7 +10,6 @@ from types import TracebackType
 from typing import TYPE_CHECKING, Any, cast
 
 import duckdb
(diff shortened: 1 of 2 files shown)


tekumara:
@DanCardin i've switched to pyarrow instead, so we don't need pandas at all now

PR #240, opened 2025-07-11
feat: support scale argument in to_timestamp function

alexxv:
Snowflake `TO_TIMESTAMP` can accept a `scale` parameter that specifies time units of the number

For seconds, scale = 0
For milliseconds, scale = 3
For microseconds, scale = 6
For nanoseconds, scale = 9

I have adjusted both `_fs_to_timestamp` macro and the `to_timestamp` transformation to take `scale` into account.
Otherwise calls like `TO_TIMESTAMP(1752253006000, 3)` would create incorrect timestamp

### DIFF
--- a/fakesnow/macros.py
+++ b/fakesnow/macros.py
@@ -44,12 +44,12 @@ FS_TO_TIMESTAMP = Template(
 CREATE OR REPLACE MACRO ${catalog}._fs_to_timestamp(val, scale) AS (
     CASE
         WHEN try_cast(val AS BIGINT) IS NOT NULL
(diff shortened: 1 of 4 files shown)


tekumara:
Thanks @alexxv! Check out the tests in test_timestamp.py - it would be good to have a test case or two for the `scale` param there.

alexxv:
@tekumara Fixed the failing tests and added a couple for `to_timestamp` with `scale` argument
btw awesome project, thanks!

tekumara:
Looks great thank you @alexxv!

Issue #235, opened 2025-07-02
Syntax error for CTAS queries with column name-type pairs


skirino:
(First of all thank you for this nice library!)

CTAS queries with `(col_name col_type, ...)` after table name is [supported by snowflake](https://docs.snowflake.com/en/sql-reference/sql/create-table#create-table-as-select-also-referred-to-as-ctas), whereas [duckdb does not support the syntax](https://duckdb.org/docs/stable/sql/statements/create_table.html#syntax). When we give such a CTAS query to fakesnow, the column list is sent to duckdb, resulting in a syntax error. [dbt-snowflake](https://github.com/dbt-labs/dbt-snowflake) heavily uses this kind of CTAS queries.

This could be an issue in [sqlglot](https://github.com/tobymao/sqlglot); if that's the case I'll post another issue upstream.

---

To reproduce:

Suppose we have the following test.py file.
```py
import fakesnow
import snowflake.connector

def exec(conn, q):
    for r in conn.execute_string(q):
        print(r.fetchone())

with fakesnow.patch():
    conn = snowflake.connector.connect()
    exec(conn, "CREATE DATABASE D;")
    exec(conn, "CREATE SCHEMA D.S;")
    exec(conn, "CREATE TABLE D.S.T1 (col1 VARCHAR);")
    exec(conn, "CREATE TABLE D.S.T2 AS SELECT * FROM D.S.T1;")
    exec(conn, "CREATE TABLE D.S.T3 (col1 VARCHAR) AS SELECT * FROM D.S.T1;")
```

Creation of table `T3` fails with a syntax error. In the logs we can see that the column list is sent unmodified.
```sh
$ FAKESNOW_DEBUG=debug python test.py
ATTACH DATABASE ':memory:' AS D;
SELECT 'Database D successfully created.' as 'status';
('Database D successfully created.',)
CREATE SCHEMA D.S;
SELECT 'Schema S successfully created.' as 'status';
('Schema S successfully created.',)
CREATE TABLE D.S.T1 (COL1 TEXT);
SELECT 'Table T1 successfully created.' as 'status';
('Table T1 successfully created.',)
CREATE TABLE D.S.T2 AS SELECT * FROM D.S.T1;
SELECT 'Table T2 successfully created.' as 'status';
('Table T2 successfully created.',)
CREATE TABLE D.S.T3 (COL1 TEXT) AS SELECT * FROM D.S.T1;
Traceback (most recent call last):
[...]

PR #243, opened 2025-07-15
Support subqueries in CTAS with column list

skirino:
Thanks for the fix of https://github.com/tekumara/fakesnow/issues/235 and the release 0.9.46!

Unfortunately a CTAS query in our project (generated by dbt-snowflake) is not converted and still results in a `Parser Error: syntax error at or near "AS"`. The query looks like:

```py
>>> sqlglot.parse_one("""create or replace table t1(id varchar) as (select * from values (1))""")
Create(
  this=Schema(
    this=Table(
      this=Identifier(this=t1, quoted=False)),
    expressions=[
      ColumnDef(
        this=Identifier(this=id, quoted=False),
        kind=DataType(this=Type.VARCHAR, nested=False))]),
  kind=TABLE,
  replace=True,
  expression=Subquery(
    this=Select(
      expressions=[
        Star()],
      from=From(
        this=Values(
          expressions=[
            Tuple(
              expressions=[
                Literal(this=1, is_string=False)])])))))
```

That is, the select query part is surrounded by parentheses and parsed into a `exp.Subquery`.

This PR adds support of such CTAS with subquery by
- change the condition for the transformation
- `unnest()` the subquery to make a `exp.Select`

My environment:
- fakesnow 0.9.46
- snowflake-connector-python 3.16.0
- sqlglot 26.33.0
- python 3.11.11
- macOS 15.5

### DIFF
--- a/fakesnow/transforms/transforms.py
+++ b/fakesnow/transforms/transforms.py
@@ -1260,7 +1260,7 @@ def create_table_as(expression: exp.Expression, duck_conn: DuckDBPyConnection) -
     if (
         isinstance(expression, exp.Create)
         and expression.kind == "TABLE"
-        and isinstance(expression.expression, exp.Select)
+        and isinstance(expression.expression, (exp.Select, exp.Subquery))
         and isinstance(expression.this, exp.Schema)
         and len(expression.this.expressions) > 0
     ):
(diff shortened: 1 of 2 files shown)


tekumara:
Thank you for the fix @skirino !

PR #248, opened 2025-07-22
fix: ALTER TABLE ADD COLUMN IF NOT EXISTS for multiple columns

ca0904:
# fix: ALTER TABLE ADD COLUMN IF NOT EXISTS for multiple columns

## Summary
Fixed the `alter_table_add_multiple_columns` function to properly handle the `IF NOT EXISTS`  clause when adding multiple columns. In Snowflake, `IF NOT EXISTS` applies to all columns  in the statement, but the function was only applying it to the first column.

## Problem
```sql
-- Input
ALTER TABLE tab1 ADD COLUMN IF NOT EXISTS col1 INT, col2 VARCHAR(100);

-- Previous incorrect output
ALTER TABLE tab1 ADD COLUMN IF NOT EXISTS col1 INT;
ALTER TABLE tab1 ADD COLUMN col2 VARCHAR(100);  -- Missing IF NOT EXISTS
```

## Solution
Enhanced the function to detect when any column has `IF NOT EXISTS` and propagate it to  all columns in the statement.

**Key change**: Fixed access to the exists flag using `action.args.get('exists', False)`  instead of `getattr(action, 'exists', False)`.

## Test Coverage
- Unit test `test_alter_table_add_multiple_columns` validates the transformation logic
- Integration test in `test_fakes.py` verifies end-to-end functionality.

## Files Changed
- `fakesnow/transforms/ddl.py` - Enhanced transformation function
- `tests/test_ddl.py` - Enhanced existing test
- `tests/test_fakes.py` - Enabled integration test

### DIFF
--- a/fakesnow/transforms/ddl.py
+++ b/fakesnow/transforms/ddl.py
@@ -19,7 +19,7 @@ from sqlglot import exp
 def alter_table_add_multiple_columns(expression: exp.Expression) -> list[exp.Expression]:
     """Transform ALTER TABLE ADD COLUMN with multiple columns into separate statements.
 
-    Snowflake supports: ALTER TABLE tab1 ADD COLUMN col1 INT, col2 VARCHAR(50), col3 BOOLEAN;
+    Snowflake supports: ALTER TABLE IF EXISTS tab1 ADD COLUMN IF NOT EXISTS col1 INT, col2 VARCHAR(50), col3 BOOLEAN;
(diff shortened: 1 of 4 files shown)


tekumara (approved):

ca0904:
what about this variable `SUCCESS_NOP`?
should I import it from `transforms.py`?

tekumara:
ah yep

PR #326, opened 2026-04-10
feat: implicit VARCHAR-to-numeric cast for aggregate functions

mattotodd:
## Problem

Snowflake implicitly casts VARCHAR columns to numeric when used in aggregate functions like SUM(), AVG(), MEDIAN(), etc. DuckDB is strict and rejects these:

```
Binder Error: No function matches the given name and argument types 'sum(VARCHAR)'.
```

This is a common pattern when data is loaded via CSV (everything comes in as VARCHAR) and then aggregated without explicit casts.

## Solution

Add a `numeric_agg_implicit_cast` transform that wraps arguments to numeric-only aggregate functions with `TRY_CAST(... AS DOUBLE)`, matching Snowflake's implicit coercion behavior.

- Only applies to numeric-only aggregates: SUM, AVG, MEDIAN, VARIANCE, STDDEV, etc.
- Does NOT affect COUNT, MIN, MAX, ARRAY_AGG, etc. which work on any type
- Skips arguments that are already Cast/TryCast expressions (no double-wrapping)

## Example

```sql
-- Input (Snowflake SQL)
SELECT SUM(amount) FROM t

-- After transform (DuckDB-compatible)
SELECT SUM(TRY_CAST(amount AS DOUBLE)) FROM t
```

## Tests

- Unit tests in `test_transforms.py` covering SUM, AVG, COUNT (unchanged), MAX (unchanged), already-cast args, and mixed queries
- Integration test in `test_fakes.py` verifying end-to-end SUM/AVG on a VARCHAR column

### DIFF
--- a/fakesnow/cursor.py
+++ b/fakesnow/cursor.py
@@ -316,6 +316,7 @@ class FakeSnowflakeCursor:
             .transform(transforms.create_clone)
             .transform(transforms.alias_in_join)
             .transform(transforms.alter_table_strip_cluster_by)
+            .transform(transforms.numeric_agg_implicit_cast)
             .transform(lambda e: transforms.create_stage(e, self._conn.database, self._conn.schema))
             .transform(lambda e: transforms.list_stage(e, self._conn.database, self._conn.schema))
             .transform(lambda e: transforms.put_stage(e, self._conn.database, self._conn.schema, params))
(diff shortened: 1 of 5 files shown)


tekumara (approved):
thanks for the contribution!

PR #333, opened 2026-05-26
feat: publish docker image

### LINKED ISSUE #201: FR: publish docker image
It would be great if `fakesnow` published a Docker image. This will be useful for several reasons:

- dependency management: given that some of fakesnow's dependencies need to be pinned, it would allow to use it in many more projects and prevent dependency clashes
- usable for non python projects
- even for Python projects it could be desirable to use a docker image, as in that case the project will still use the connector library. It would also allow to use other query libraries like snowpark

gao-artur:
Fix #201 

Hey. I added a Dockefile to the project and updated the README file. I assumed the image would be published under `ghcr.io/tekumara/fakesnow`, but let me know if you prefer another name.
Also added an integration test using [testcontainers](https://github.com/testcontainers/testcontainers-python), which builds the Docker image, starts the container, and runs a simple query to ensure the server starts and runs the `fakesnow` server.
I also built and started the container locally and ran my .NET test suite, so I can confirm that everything is fine.

**Disclaimer**: I have zero knowledge of Python. The code was written by Copilot, but not in autopilot mode.


Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+WORKDIR /app
+
+COPY . .
+
+RUN pip install --no-cache-dir ".[server]"
+
+EXPOSE 8080
+
+CMD ["fakesnow", "-s", "--host", "0.0.0.0", "--port", "8080"]
```
tekumara: this is the easy part :-) what is really needed is a workflow to publish this to ghcr

integration_tests/test_docker.py
```diff
@@ -0,0 +1,40 @@
...
+    with DockerImage(path=str(REPO_ROOT), tag="fakesnow-test:latest") as image:
+        with (
+            DockerContainer(str(image))
+            .with_exposed_ports(PORT)
+            .waiting_for(LogMessageWaitStrategy("Application startup complete").with_startup_timeout(30))
+        ) as container:
+            yield container
+
+
+def test_docker_select(docker_fakesnow: DockerContainer) -> None:
```
tekumara: i'd prefer to keep the unit tests quick and free of external dependencies (like having a running docker daemon) ...  ..  although something like this could be part of the docker build pipeline and use the snowflake cli to smoke test the image

Dockerfile
gao-artur: Sure, I'll check how to do that and will update the PR. Hopefully today or tomorrow.

gao-artur: Btw, why ghcr and not dockerhub? I think the latter is more convenient for OSS projects

integration_tests/test_docker.py
gao-artur: I moved this test to `integration_tests` folder and ran it on `publish-docker` action. If this pattern sounds correct, should the `tests` folder be renamed to `unit_tests` for consistency?

Dockerfile
tekumara: ghcr is friendlier to users because it doesn't have pull rate limits like dockerhub does

.github/workflows/release.yml
```diff
@@ -35,3 +35,48 @@ jobs:
...
+          key: ${{ runner.os }}-py${{ steps.setup-python.outputs.python-version }}-venv-${{ hashFiles('pyproject.toml') }}
+      - if: steps.cache-venv.outputs.cache-hit != 'true'
+        run: make install
+      - name: Test Docker image
+        run: uv run pytest integration_tests/test_docker.py
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
```
gao-artur: I think publishing to dockerhub is preferable to using ghcr because it doesn't require authentication to `docker pull` and it's also easier discoverable (`tekumara/fakesnow` vs `ghcr.io/tekumara/fakesnow`).
If you are agree, then you need to configure the secrets:
- Log in to your Docker Hub account.
- Navigate to Account Settings -> Personal access tokens.
- Generate new token
- Select Red & Write permissions

.github/workflows/release.yml
```diff
@@ -35,3 +35,48 @@ jobs:
...
+      - name: Extract Docker metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: tekumara/fakesnow
+          tags: |
+            type=semver,pattern={{version}}
+            type=raw,value=latest,enable=${{ !github.event.release.prerelease }}
```
gao-artur: The `latest` tag is not added to a prerelease versions

Dockerfile
gao-artur: But it requires authentication to `docker pull`. IMO not the best dev experience. Also, the rate limit is only important for CI, local dev will probably never hit the rate limit. And authenticating with a free account in CI doubles the rate limit: from 100 requests per 6 hours for anonymous pulls to 200 requests per 6 hours for authenticated pulls. Of course, if you don't agree, I'll change it to ghcr.

gao-artur: Maybe I'm wrong and public images don't require authentication. Checking.

gao-artur: Done. Updated the publish to `ghcr.io/tekumara/fakesnow`

README.md
```diff
@@ -90,6 +90,12 @@ with fakesnow.patch(db_path="databases/"):
 
 For scenarios where patching won't work (like subprocesses or non-Python clients), you can run fakesnow as an HTTP server.
 
+#### Using Docker
+
+```shell
+docker run -p 8080:8080 tekumara/fakesnow:latest
```
gao-artur:
```suggestion
docker run -p 8080:8080 ghcr.io/tekumara/fakesnow:latest
```

.github/workflows/release.yml
```diff
@@ -35,3 +35,51 @@ jobs:
...
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Extract Docker metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ghcr.io/gao-artur/fakesnow
```
gao-artur:
```suggestion
          images: ghcr.io/tekumara/fakesnow
```

gao-artur:
@tekumara is anything missing here?

tekumara:
thanks for the contribution @gao-artur! 
ps: i made a few changes including using chainguard base images

PR #334, opened 2026-05-28
feat: Add named params support in server mode

gao-artur:
I tried running parameterized queries with named params with the .NET connector, but failed because apparently, server mode only supports positional params. Added named params support in this PR.

**Disclaimer**: I have zero knowledge of Python. The code was written by Copilot. Let me know if something can be improved.

### DIFF
--- a/fakesnow/copy_into.py
+++ b/fakesnow/copy_into.py
@@ -16,7 +16,7 @@ from sqlglot import Expr, exp
 
 import fakesnow.transforms.stage as stage
 from fakesnow import logger
-from fakesnow.params import MutableParams, pop_qmark_param
+from fakesnow.params import MutableParams, get_named_param, pop_qmark_param
 
 Params = Sequence[Any] | dict[Any, Any]
 
@@ -237,7 +237,7 @@ def _params(expr: exp.Copy, params: MutableParams | None = None) -> CopyParams:
             if isinstance(param.expression, exp.Var):
                 on_error = param.expression.name.upper()
             elif isinstance(param.expression, exp.Placeholder):
-                on_error = pop_qmark_param(params, expr, param.expression)
+                on_error = get_named_param(params, param.expression) if param.expression.this else pop_qmark_param(params, expr, param.expression)
             else:
                 raise NotImplementedError(f"{param.expression.__class__=}")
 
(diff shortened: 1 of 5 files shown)


fakesnow/copy_into.py
```diff
@@ -237,7 +237,7 @@ def _params(expr: exp.Copy, params: MutableParams | None = None) -> CopyParams:
             if isinstance(param.expression, exp.Var):
                 on_error = param.expression.name.upper()
             elif isinstance(param.expression, exp.Placeholder):
-                on_error = pop_qmark_param(params, expr, param.expression)
+                on_error = get_named_param(params, param.expression) if param.expression.this else pop_qmark_param(params, expr, param.expression)
```
tekumara: is there a test to cover this? otherwise I suggest we remove this change and the one in transforms.py and params.py

tests/test_server.py
```diff
@@ -77,6 +77,54 @@ def test_server_binding_qmark(server: dict):
...
+    )
+    assert result["success"] is True
+    assert result["data"]["total"] == 1
+
+    # no match
+    result = query("select * from customers where id = :myid", bindings={"myid": {"type": "FIXED", "value": "99"}})
+    assert result["success"] is True
+    assert result["data"]["total"] == 0
+
+
```
tekumara, line 92: can be simplified to:
```suggestion
def test_server_binding_named(server: dict) -> None:
    with snowflake.connector.connect(**server, database="DB1", schema="SCHEMA1") as conn:

        with conn.cursor() as cur:
            cur.execute("create table customers (id int, name text)")
            cur.execute("insert into customers values (1, 'Jenny'), (2, 'Jasper')")

        def query(sql: str, bindings: dict[str, dict[str, str]]) -> dict:
            # Use cmd_query to send the named bindings to the server.
            # With the default pyformat paramstyle, cursor.execute client-side interpolates named
            # params instead of sending them to the server
            return conn.cmd_query(sql, conn._next_sequence_counter(), uuid.uuid4(), binding_params=bindings)  # noqa: SLF001

        result = query(
            "select * from customers where id = :myid and name = :myname",
            {"myid": {"type": "FIXED", "value": "2"}, "myname": {"type": "TEXT", "value": "Jasper"}},
        )
        assert result["success"] is True
        assert result["data"]["total"] == 1
```
(this also makes it easier to verify against a real snowflake instance)

tekumara (commented):
thanks for raising this! some small suggestions below

fakesnow/copy_into.py
gao-artur: As I said, I don't know python, so this PR was fully written by Copilot. If you think this change is not necessary, I'll revert it.

PR #344, opened 2026-06-26
fix: handle CREATE MACRO in server mode to avoid ProgrammingError

### LINKED ISSUE #343: Server mode raises ProgrammingError on CREATE MACRO despite creating the macro
## Summary

In **server mode**, executing `CREATE MACRO` raises a `ProgrammingError` even though the macro is created successfully. The same statement works fine in library mode.

```
snowflake.connector.errors.ProgrammingError: 001003 (42000): Parser Error: syntax error at or near "CREATE"
LINE 1: DESCRIBE CREATE OR REPLACE MACRO ...
                 ^
```

This bites anyone who registers macros over the wire — e.g. re-declaring the helpers in `macros.py` (`_fs_to_timestamp`, `_fs_flatten`) against a running server, which library mode auto-registers per-catalog but server mode does not.

## Root cause

`CREATE MACRO` is DuckDB-only DDL, so sqlglot can't parse it and falls back to an opaque `Command(this="CREATE", expression=" OR REPLACE MACRO ...")`. In `FakeSnowflakeCursor._execute`:

- `expr.key_command()` returns just `"CREATE"`, and `transformed.find(exp.Identifier)` is `None`, so it matches **none** of the `result_sql` branches (CREATE TABLE/VIEW/SCHEMA/SECRET/…).
- With `result_sql` left `None`, `self._last_sql = result_sql or sql` keeps the **raw `CREATE OR REPLACE MACRO …` DDL**. Every other DDL gets rewritten to a `SELECT '… successfully created.' as status` row.
- The macro itself is created fine (`self._duck_conn.execute(sql)` succeeds).

Then in `server.py`, server mode describes the result **unconditionally**:

```python
cur = await run_in_threadpool(conn.cursor().execute, sql_text, binding_params=params, server=True)
rowtype = describe_as_rowtype(cur._describe_last_sql())
```

`_describe_last_sql()` runs `DESCRIBE <last_sql>` → `DESCRIBE CREATE OR REPLACE MACRO …`, which DuckDB rejects with a parser error. Library mode is unaffected because it only describes lazily via `.description`.

## Repro (mirrors the server.py path)

```python
import fakesnow, snowflake.connector

with fakesnow.patch():
    conn = snowflake.connector.connect(database="DB1", schema="SCHEMA1")
    cur = conn.cursor()
    for s in ("CREATE DATABASE IF NOT EXISTS DB1", "CREATE SCHEMA IF NOT EXISTS DB1.SCHEMA1",
              "USE DATABASE DB1", "USE SCHEMA SCHEMA1"):
        cur.execute(s)

    cur.execute("CREATE OR REPLACE MACRO inc(x) AS (x + 1)", server=True)
    print(cur._last_sql)            # raw "CREATE OR REPLACE MACRO ..." DDL
    cur._describe_last_sql()        # ProgrammingError: syntax error at or near "CREATE"

    cur.execute("SELECT inc(41)")   # but the macro IS created:
    print(cur.fetchall())           # [(42,)]
```

(Reproduced on 0.11.9 and 0.11.5.)

## Suggested fix

Give `CREATE MACRO` a status-row `result_sql` like the other DDL, so `_last_sql` becomes a describable `SELECT`. A branch alongside the existing `SECRET` handler in `_execute` does it:

```python
elif (
    cmd == "CREATE"
    and isinstance(transformed, exp.Command)
    and isinstance(transformed.expression, str)
    and re.search(r"\bMACRO\b", transformed.expression, re.IGNORECASE)
):
    result_sql = SQL_SUCCESS
```

Happy to send a PR — I have this change plus a `test_server_create_macro` regression test ready.

mdz:
## Summary

- `CREATE MACRO` is DuckDB-only DDL that sqlglot parses as an opaque `Command`, so it matched none of the `result_sql` branches in `FakeSnowflakeCursor._execute` and left `_last_sql` as the raw DDL string.
- Server mode calls `DESCRIBE <_last_sql>` unconditionally, so `DESCRIBE CREATE OR REPLACE MACRO …` hit DuckDB's parser and raised a `ProgrammingError`, even though the macro was created successfully.
- Fix: add a branch in `_execute` that detects `CREATE … MACRO` (via regex on the opaque `Command.expression`) and assigns `SQL_SUCCESS` as the `result_sql`, consistent with how other DDL is handled.

Fixes #343

## Changes

- [`fakesnow/cursor.py`](fakesnow/cursor.py): new `elif` branch alongside the `SECRET` handler for `CREATE MACRO`
- [`tests/test_server.py`](tests/test_server.py): `test_server_create_macro` regression test

## Test plan

- [x] `uv run pytest tests/test_server.py::test_server_create_macro` passes
- [x] Existing server tests unaffected

### DIFF
--- a/tests/test_describe.py
+++ b/tests/test_describe.py
@@ -315,3 +315,9 @@ def test_description_update(dcur: snowflake.connector.cursor.DictCursor):
         ResultMetadata(name='number of multi-joined rows updated', type_code=0, display_size=None, internal_size=None, precision=38, scale=0, is_nullable=True)
     ]
     # fmt: on
+
+
+def test_description_create_macro(dcur: snowflake.connector.cursor.DictCursor):
+    dcur.execute("create or replace macro inc(x) as (x + 1)")
+    assert dcur.fetchall() == [{"status": "Statement executed successfully."}]
+    assert dcur.description == [ResultMetadata(name='status', type_code=2, display_size=None, internal_size=16777216, precision=None, scale=None, is_nullable=True)]  # fmt: skip
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -419,14 +419,6 @@ def test_server_response_params(server: dict) -> None:
     assert {"name": "AUTOCOMMIT", "value": False} in response.json()["data"]["parameters"]
 
 
-def test_server_create_macro(scur: snowflake.connector.cursor.SnowflakeCursor) -> None:
-    cur = scur
-    # CREATE MACRO is DuckDB-only DDL; server mode must not raise ProgrammingError
-    cur.execute("CREATE OR REPLACE MACRO inc(x) AS (x + 1)")
-    cur.execute("SELECT inc(41)")
-    assert cur.fetchall() == [(42,)]
-
-
 def test_server_rowcount(scur: snowflake.connector.cursor.SnowflakeCursor):
     cur = scur
 
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -419,6 +419,14 @@ def test_server_response_params(server: dict) -> None:
     assert {"name": "AUTOCOMMIT", "value": False} in response.json()["data"]["parameters"]
 
 
+def test_server_create_macro(scur: snowflake.connector.cursor.SnowflakeCursor) -> None:
+    cur = scur
+    # CREATE MACRO is DuckDB-only DDL; server mode must not raise ProgrammingError
+    cur.execute("CREATE OR REPLACE MACRO inc(x) AS (x + 1)")
+    cur.execute("SELECT inc(41)")
+    assert cur.fetchall() == [(42,)]
+
+
 def test_server_rowcount(scur: snowflake.connector.cursor.SnowflakeCursor):
     cur = scur
 
--- a/fakesnow/cursor.py
+++ b/fakesnow/cursor.py
@@ -528,6 +528,13 @@ class FakeSnowflakeCursor:
             match = re.search(r"SECRET\s+(\w+)\s*\(", transformed.expression, re.IGNORECASE)
             secret_name = match[1].upper() if match else "UNKNOWN"
             result_sql = SQL_CREATED_SECRET.substitute(name=secret_name)
+        elif (
+            cmd == "CREATE"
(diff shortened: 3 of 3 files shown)


tests/test_server.py
```diff
@@ -419,6 +419,14 @@ def test_server_response_params(server: dict) -> None:
     assert {"name": "AUTOCOMMIT", "value": False} in response.json()["data"]["parameters"]
 
 
+def test_server_create_macro(scur: snowflake.connector.cursor.SnowflakeCursor) -> None:
```
tekumara: Can we  replace this with a test similar to this:

https://github.com/tekumara/fakesnow/blob/8bf050eac5dd676e0810d5facd3571c57bf873e5/tests/test_describe.py#L250

tekumara (commented):
Thanks for this! Just added a suggestion for a test more consistent with the existing tests

tests/test_describe.py
```diff
@@ -315,3 +315,9 @@ def test_description_update(dcur: snowflake.connector.cursor.DictCursor):
         ResultMetadata(name='number of multi-joined rows updated', type_code=0, display_size=None, internal_size=None, precision=38, scale=0, is_nullable=True)
     ]
     # fmt: on
+
+
+def test_description_create_macro(dcur: snowflake.connector.cursor.DictCursor):
+    dcur.execute("create or replace macro inc(x) as (x + 1)")
```
tekumara, line 321: it just occurred to me that snowflake doesn't support macros, and you can't run this on a real snowflake instance... so i'm hesitant to support it ... what is the use case for creating macros?

mdz: I've used it to implement support for Snowflake functions that fakesnow doesn't support. But if you prefer, I can submit patches to add first-class support for those instead. Would that be better?

mdz: also, FYI, CREATE MACRO already works [it defines the macro], but emits a spurious error return. if you want to make it fail cleanly, that will require a code change

mdz: I submitted separate issues and PRs for the problems I was working around with CREATE MACRO, along with a PR to reject CREATE MACRO cleanly. We can close this one?

mdz:
Closing in favor of #350

PR #349, opened 2026-06-29
fix: create macros for runtime-created databases

### LINKED ISSUE #347: Server mode: internal _fs_* macros not registered for databases created via CREATE DATABASE
## Summary

In server mode, the internal `_fs_*` macros that sqlglot's transforms depend on (`_fs_to_timestamp`, `_fs_flatten`, and presumably `_fs_object_construct`) are not registered for databases that are created via a runtime `CREATE DATABASE` statement. As a result, queries that get rewritten to call those macros fail with errors like:

```
Scalar Function with name _fs_to_timestamp does not exist!
Table Function with name _fs_flatten does not exist!
```

even though the equivalent queries succeed in library mode.

## Likely root cause

`macros.creation_sql(catalog)` is only invoked from the `create_database` branch of `FakeSnowflakeConnection.__init__`:

```python
# fakesnow/conn.py
duck_conn.execute(info_schema.per_db_creation_sql(self.database))
duck_conn.execute(macros.creation_sql(self.database))
```

So the `_fs_*` macros are registered **only** for databases that exist (or are named on connect) at connection-construction time. A database created later by executing `CREATE DATABASE foo` over an existing connection never has `macros.creation_sql("foo")` run against it.

This is invisible in typical library-mode usage (you connect with the target database, so its macros get created), but it bites server mode: a client connects once against `shared_fs`, then issues `CREATE DATABASE ...` statements for each catalog it needs. Those catalogs are missing the `_fs_*` macros, so the first `TO_TIMESTAMP_NTZ(...)` / `LATERAL FLATTEN(INPUT => ...)` against them fails.

## Reproduction

Against fakesnow server mode (observed via the Snowflake Go client, but the mechanism is client-agnostic):

1. Connect.
2. `CREATE DATABASE foo;` then use it.
3. `SELECT TO_TIMESTAMP_NTZ('1672531200', 0);` → `_fs_to_timestamp does not exist`.
4. `SELECT ... FROM t, LATERAL FLATTEN(INPUT => t.arr_col);` → `_fs_flatten does not exist`.

The same statements work in library mode where the database was passed at connect time.

## Workaround

We re-create the macros ourselves as catalog-less macros in our seed SQL, copied verbatim from `fakesnow/macros.py` (minus the `${catalog}.` prefix). This unblocks both `TO_TIMESTAMP_NTZ` and structured-array `LATERAL FLATTEN` in server mode. (Note: this requires #344 first, otherwise the `CREATE MACRO` statements themselves raise a `ProgrammingError` in server mode due to the DESCRIBE-prepending behavior.)

## Suggested fix

Register the `_fs_*` macros for any database when it's created, not only in the connection constructor — e.g. run `macros.creation_sql(catalog)` (and `info_schema.per_db_creation_sql`) on `CREATE DATABASE` execution as well, so runtime-created catalogs get the same treatment as connect-time ones.

## Context

Surfaced while running DoorDash's analytics queries against fakesnow in server mode. Related: #343 / #344 (the `CREATE MACRO` DESCRIBE error that must be fixed for the catalog-less workaround above to load).

mdz:
### Why?

Fixes #347.

Fixes a DuckDB catalog error where Snowflake-compatible timestamp conversion can fail after creating a database with `CREATE DATABASE`:

`Scalar Function with name _fs_to_timestamp does not exist`

This change is needed so databases created during a fakesnow session support the same internal compatibility helpers as databases created at connection time.

### How?

Initialize newly created databases with fakesnow’s internal helper macros, so Snowflake SQL rewrites continue to work consistently after switching into a database created during the session.

<sub>Generated with Claude Code</sub>

### DIFF
--- a/fakesnow/cursor.py
+++ b/fakesnow/cursor.py
@@ -26,6 +26,7 @@ from typing_extensions import Self
 import fakesnow.checks as checks
 import fakesnow.expr as expr
 import fakesnow.info_schema as info_schema
+import fakesnow.macros as macros
 import fakesnow.transforms as transforms
 from fakesnow import logger
 from fakesnow.copy_into import copy_into
@@ -446,6 +447,7 @@ class FakeSnowflakeCursor:
         elif create_db_name := transformed.args.get("create_db_name"):
             # we created a new database, so create the info schema extensions
             self._duck_conn.execute(info_schema.per_db_creation_sql(create_db_name))
+            self._duck_conn.execute(macros.creation_sql(create_db_name))
             result_sql = SQL_CREATED_DATABASE.substitute(name=create_db_name)
 
         elif stage_name := transformed.args.get("create_stage_name"):
--- a/tests/test_connect.py
+++ b/tests/test_connect.py
@@ -90,21 +90,20 @@ def test_connect_db_path_can_create_database() -> None:
 def test_create_database_registers_macros() -> None:
     # _fs_* macros must be registered for databases created via CREATE DATABASE,
     # not only for databases passed at connect time (issue #347)
-    with fakesnow.patch():
-        with snowflake.connector.connect() as conn:
-            cur = conn.cursor()
-            cur.execute("CREATE DATABASE foo")
-            cur.execute("USE DATABASE foo")
-            cur.execute("CREATE SCHEMA foo.bar")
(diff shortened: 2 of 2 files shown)


tekumara:
thank you!

PR #350, opened 2026-06-29
fix: raise ProgrammingError for CREATE MACRO instead of creating it as a side effect

mdz:
## What changed

`CREATE MACRO` is DuckDB-only DDL that Snowflake doesn't support. Previously fakesnow would silently pass it through to DuckDB, actually creating the macro as a side effect (and in server mode, also raising a `ProgrammingError` when result post-processing tried to `DESCRIBE` the raw DDL string).

Per the reviewer's position in #344, the right behavior is for `CREATE MACRO` to fail cleanly rather than succeeding with a DuckDB-only side effect.

## How

Added an early guard in `FakeSnowflakeCursor._execute` ([`fakesnow/cursor.py`](fakesnow/cursor.py)) that detects `CREATE ... MACRO` (sqlglot parses this as an opaque `exp.Command`) and raises a `ProgrammingError` with `errno=2 / sqlstate=0A000` (feature not supported) before DuckDB ever executes it.

## Test plan

- `CREATE OR REPLACE MACRO` raises `ProgrammingError: SQL compilation error: Unsupported feature 'CREATE MACRO'.`
- The macro is **not** created in DuckDB — calling the macro name after the error also raises an error
- Existing cursor tests pass (`uv run pytest tests/test_cursor.py`)

### DIFF
--- a/fakesnow/cursor.py
+++ b/fakesnow/cursor.py
@@ -355,6 +355,17 @@ class FakeSnowflakeCursor:
         if not sql:
             raise NotImplementedError(transformed.sql(dialect="snowflake"))
 
+        if (
+            cmd == "CREATE"
+            and isinstance(transformed, exp.Command)
+            and re.search(r"\bMACRO\b", transformed.expression, re.IGNORECASE)
+        ):
+            raise snowflake.connector.errors.ProgrammingError(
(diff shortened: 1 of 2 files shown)


tekumara:
thank you for this! 

ps: i've adjusted the error message to match real snowflake

PR #365, opened 2026-08-09
fix: return statementTypeId so DML and DDL statements work when using the JDBC driver

### LINKED ISSUE #204: Statement 'USE SCHEMA SCHEMA2' cannot be executed using current API
When changing the schema in Dbeaver via Choose Database/Schema (`CTRL+0`) the JDBC driver errors with:
```
SQL Error [200042] [0A000]: Statement 'USE SCHEMA SCHEMA2' cannot be executed using current API.
Statement 'USE SCHEMA SCHEMA2' cannot be executed using current API.
Statement 'USE SCHEMA SCHEMA2' cannot be executed using current API.
```

From _~/Library/DBeaverData/workspace6/.metadata/dbeaver-debug.log_:

```
2025-03-24 16:04:18.763 - Error during job 'Change active database' execution
net.snowflake.client.jdbc.SnowflakeSQLLoggedException: Statement 'USE SCHEMA SCHEMA2' cannot be executed using current API.
        at net.snowflake.client.jdbc.SnowflakeStatementV1.executeUpdateInternal(SnowflakeStatementV1.java:256)
        at net.snowflake.client.jdbc.SnowflakeStatementV1.executeLargeUpdate(SnowflakeStatementV1.java:206)
        at net.snowflake.client.jdbc.SnowflakeStatementV1.executeUpdate(SnowflakeStatementV1.java:192)
        at org.jkiss.dbeaver.model.impl.jdbc.exec.JDBCStatementImpl.executeUpdate(JDBCStatementImpl.java:375)
        at org.jkiss.dbeaver.ext.snowflake.model.SnowflakeExecutionContext.setActiveSchema(SnowflakeExecutionContext.java:201)
        at org.jkiss.dbeaver.ext.snowflake.model.SnowflakeExecutionContext.setDefaultSchema(SnowflakeExecutionContext.java:131)
        at org.jkiss.dbeaver.ext.snowflake.model.SnowflakeExecutionContext.setDefaultSchema(SnowflakeExecutionContext.java:123)
        at org.jkiss.dbeaver.ext.generic.model.GenericExecutionContext.setDefaultSchema(GenericExecutionContext.java:1)
        at org.jkiss.dbeaver.model.exec.DBExecUtils.changeDefaultObject(DBExecUtils.java:452)
        at org.jkiss.dbeaver.model.exec.DBExecUtils.setExecutionContextDefaults(DBExecUtils.java:411)
        at org.jkiss.dbeaver.ui.actions.datasource.SelectActiveSchemaHandler$2.run(SelectActiveSchemaHandler.java:214)
        at org.jkiss.dbeaver.model.runtime.AbstractJob.run(AbstractJob.java:119)
        at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

See https://github.com/snowflakedb/snowflake-jdbc/blob/722137cd34bb87a85662981dfd02464ab702aaae/src/main/java/net/snowflake/client/jdbc/SnowflakeStatementV1.java#L251

anjeongkyun:
Without `statementTypeId` the JDBC driver falls back to `SFStatementType.UNKNOWN`, which is flagged as generating a result set, so `executeUpdate` is rejected for every statement — DDL, DML and `USE` alike. The statement still runs, so rows get written and the caller sees an exception anyway. That blocks `JdbcTemplate.update()`, Hibernate and Flyway.

Adding the id on its own regresses the python connector: once it sees a DML type id it stops using `total` for rowcount and reads affected counts out of `rowset`, which we weren't sending because DML results went back as arrow. DML results are now returned as json instead, which is what real Snowflake does (checked against an account — details in #204). That also corrects `merge`, which reported the result row count (1) rather than rows inserted + updated (2).

Checked with snowflake-jdbc 4.3.2 against the server:

| | before | after |
|---|---|---|
| `executeUpdate` `INSERT ... VALUES (2),(3)` | exception | 2 |
| `executeUpdate` `UPDATE` / `DELETE` | exception | 1 |
| `executeUpdate` `USE SCHEMA` / DDL | exception | 0 |
| `PreparedStatement.executeUpdate` | exception | 1 |

Tests: statement type mapping unit tests, plus a server rowcount test covering insert/update/delete/merge whose values match a real Snowflake instance.

Fixes #204

### DIFF
--- a/fakesnow/server.py
+++ b/fakesnow/server.py
@@ -22,6 +22,7 @@ from fakesnow.expr import normalise_ident
 from fakesnow.fakes import FakeSnowflakeConnection
 from fakesnow.instance import FakeSnow
 from fakesnow.rowtype import describe_as_rowtype
+from fakesnow.statement_type import DML_TYPE_IDS, statement_type_id
 
 logger = logging.getLogger("fakesnow.server")
 # use same format as uvicorn
@@ -102,6 +103,7 @@ async def query_request(request: Request) -> JSONResponse:
             params = None
 
         expr = parse_one(sql_text, read="snowflake")
+        type_id = statement_type_id(expr)
 
         try:
             # only a single sql statement is sent at a time by the python snowflake connector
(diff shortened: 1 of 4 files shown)


tekumara (approved):
great work, thank you!

PR #369, opened 2026-08-10
docs: how to connect with the JDBC driver

anjeongkyun:
The JDBC driver works against the server but the README doesn't mention it, and there are two non-obvious requirements that cost time to work out:

- `account` has to be in the URL. The driver derives it from the host otherwise, which fails for `localhost` and produces a confusing "claims to not accept jdbcUrl" error rather than a connection failure.
- The JVM needs `--add-opens=java.base/java.nio=ALL-UNNAMED` or the bundled Arrow throws `ExceptionInInitializerError` when the first result comes back.

Both came up wiring a Spring Boot service's integration tests to fakesnow via Testcontainers, replacing a Postgres substitute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,30 @@ SELECT 'Hello World' as greeting
 +-------------+
 ```
 
+#### Connecting from Java
+
+The official [snowflake-jdbc](https://github.com/snowflakedb/snowflake-jdbc) driver connects to the server unmodified:
+
+```
+jdbc:snowflake://<host>:<port>/?ssl=off&account=fakesnow&db=<database>&schema=<schema>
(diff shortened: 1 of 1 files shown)


anjeongkyun:
Unrelated to the change itself, but it came out of the same work: I put the Testcontainers side into a small module — https://github.com/anjeongkyun/testcontainers-fakesnow. Apache-2.0, just assembles the url and carries the defaults described here.

Mentioning it so you know it exists rather than to ask for anything. Happy to rename if having `fakesnow` in the artifact name is a problem.

tekumara (approved):
thank you!

PR #375, opened 2026-08-10
fix: support JDBC driver setTimestamp

### LINKED ISSUE #374: setTimestamp() fails with 500 for all timestamp columns: from_binding only converts TIMESTAMP_NTZ
`PreparedStatement.setTimestamp()` fails for every timestamp column type over the server. Literals are fine, only bindings break.

```java
stmt.execute("CREATE OR REPLACE TABLE t (v TIMESTAMP_NTZ)");
try (PreparedStatement ps = conn.prepareStatement("INSERT INTO t VALUES (?)")) {
    ps.setTimestamp(1, Timestamp.valueOf("2026-01-01 10:00:00"));
    ps.executeUpdate();   // HTTP status=500
}
```

| bound type | fakesnow | real Snowflake |
|---|---|---|
| `TIMESTAMP_NTZ` | 500 | `2025-12-31 17:00:00.000` |
| `TIMESTAMP_LTZ` | 500 | `2025-12-31 17:00:00.000 -0800` |
| `TIMESTAMP_TZ` | 500 | `2025-12-31 17:00:00.000 -0800` |
| `TIMESTAMP` | 500 | — |
| `DATE` (`setDate`) | ok | — |
| `TIME` (`setTime`) | ok | matches (`01:00:00` both) |

Server log:

```
sql_text='INSERT INTO t VALUES (?)' params=('1767229200000000000',) Unhandled exception
_duckdb.ConversionException: Conversion Error: timestamp field value out of range: "1767229200000000000"
```

The cause is the binding type name. The JDBC driver sends `setTimestamp` as **`TIMESTAMP_LTZ`**, and `from_binding` only converts `TIMESTAMP_NTZ`:

https://github.com/tekumara/fakesnow/blob/main/fakesnow/converter.py

```python
elif type_ == "TIMESTAMP_NTZ":
    return from_datetime(value)
else:
    # For other types, return str
    return value
```

so the epoch-nanoseconds string reaches duckdb unconverted. The python connector sends `TIMESTAMP_NTZ`, which is presumably why this hasn't come up.

Since it's a 500 the driver treats it as retryable and repeats the insert several times before failing.

This blocks writing timestamps through Hibernate, Spring Data JDBC or plain `JdbcTemplate.update(...)` — anything that binds rather than inlines. Happy to send a PR adding `TIMESTAMP_LTZ` and `TIMESTAMP_TZ` to the conversion; the offset semantics of those two are a separate question (#372).

anjeongkyun:
`setTimestamp()` failed with a 500 for every timestamp column, because `from_binding` only converted `TIMESTAMP_NTZ`. The JDBC driver sends `setTimestamp` as `TIMESTAMP_LTZ`, so the epoch-nanoseconds string fell through to the `else` branch and reached duckdb unconverted:

```
_duckdb.ConversionException: Conversion Error: timestamp field value out of range: "1767229200000000000"
```

`TIMESTAMP_TZ` needed handling separately — the connector encodes it as `<epoch nanoseconds> <utc offset in minutes + 1440>`, so `int()` on the whole value fails.

Verified with snowflake-jdbc 3.19.0 against the server:

```
                 before   after
TIMESTAMP_NTZ    500      2026-01-01 01:00:00 Z
TIMESTAMP_LTZ    500      2026-01-01 01:00:00 Z
TIMESTAMP_TZ     500      2026-01-01 01:00:00 Z
TIMESTAMP        500      2026-01-01 01:00:00 Z
```

That's the same instant a real account returns for the same binding — `2025-12-31 17:00:00.000 -0800` on a session in `America/Los_Angeles` — only the rendering timezone differs.

`setDate` and `setTime` were already fine, and `setTime` matches real Snowflake including the timezone shift.

Fixes #374

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -88,7 +88,8 @@ def test_server_binding_timestamp_ltz(server: dict) -> None:
 
         cur.execute("insert into example values (?)", (("TIMESTAMP_LTZ", timestamp),))
 
-        assert cur.execute("select value from example").fetchone() == (timestamp,)
+        cur.execute("select value from example")
+        assert cur.fetchone() == (timestamp,)
 
 
 def test_server_binding_named(server: dict) -> None:
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -78,6 +78,19 @@ def test_server_binding_qmark(server: dict):
         cur.execute("select * from example where xint = ?", (1,))
 
 
+def test_server_binding_timestamp_ltz(server: dict) -> None:
+    with (
+        snowflake.connector.connect(**server, database="db1", schema="schema1", paramstyle="qmark") as conn,
+        conn.cursor() as cur,
+    ):
+        cur.execute("create or replace table example (value timestamp_ntz)")
+        timestamp = datetime.datetime(2026, 1, 1, 10)
+
+        cur.execute("insert into example values (?)", (("TIMESTAMP_LTZ", timestamp),))
+
+        assert cur.execute("select value from example").fetchone() == (timestamp,)
+
+
 def test_server_binding_named(server: dict) -> None:
     with snowflake.connector.connect(**server, database="DB1", schema="SCHEMA1") as conn:
         with conn.cursor() as cur:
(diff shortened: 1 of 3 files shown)


tests/test_server.py
```diff
@@ -78,6 +78,19 @@ def test_server_binding_qmark(server: dict):
         cur.execute("select * from example where xint = ?", (1,))
 
 
+def test_server_binding_timestamp_ltz(server: dict) -> None:
```
tekumara, line 81: @anjeongkyun FYI i've added a integration test to cover this

tekumara (approved):

Issue #372, opened 2026-08-10
timestamp_tz literal with an offset returns 500, causing client retry storms


anjeongkyun:
An offset in a `timestamp_tz` literal returns 500. Because it's a 500 the client treats it as retryable — I counted 333 retries from the python connector before I killed the test, and the JDBC driver retries 7 times.

```python
cur.execute("create or replace table tz (v timestamp_tz)")
cur.execute("insert into tz values ('2026-01-01 10:00:00')")         # ok
cur.execute("insert into tz values ('2026-01-01 10:00:00 +09:00')")  # 500
```

Server log:

```
_duckdb.ConversionException: Conversion Error: Unknown TimeZone '+09:00'!
Candidate time zones: "GMT0"
LINE 1: INSERT INTO TZ VALUES ('2026-01-01 10:00:00 +09:00')
```

DuckDB reads the offset as a time zone name. I realise a per-value offset can't round-trip on DuckDB's `TIMESTAMPTZ`, so matching Snowflake exactly may not be on the table. But surfacing it as a `ProgrammingError` — or normalising the offset to UTC on the way in — would avoid the retry storm, which is what makes this expensive in a test suite.

tekumara:
Ideally we'd match snowflake's behaviour here. What does it do?

anjeongkyun:
Checked against a real account (Snowflake 10.28.101, session timezone `America/Los_Angeles`).

**Numeric offsets are accepted, and preserved per value:**

| literal | result |
|---|---|
| `'2026-01-01 10:00:00 +09:00'` | `FixedOffset(540)` |
| `'2026-01-01 10:00:00 -0700'` | `FixedOffset(-420)` |
| `'2026-01-01 10:00:00Z'` | `UTC` |
| `'2026-01-01T10:00:00+09:00'` | `FixedOffset(540)` |
| `'2026-01-01 10:00:00'` (none) | `FixedOffset(-480)`, ie: the session timezone |

Stored in a `timestamp_tz` column it survives as written:

```
inserted                        to_varchar(v)                    extract(timezone_hour from v)
'2026-01-01 10:00:00 +09:00' ->  2026-01-01 10:00:00.000 +0900    9
'2026-01-01 10:00:00 -0700' ->  2026-01-01 10:00:00.000 -0700   -7
'2026-01-01 10:00:00'       ->  2026-01-01 10:00:00.000 -0800   -8
```

**Named zones are rejected**, which I didn't expect:

```
[...]

PR #376, opened 2026-08-11
fix: accept a space before a utc offset in timestamp literals

anjeongkyun:
Two parts, both from #372.

**The space.** Snowflake takes `'2026-01-01 10:00:00 +09:00'`, duckdb reads the offset as a time zone name and raises `Unknown TimeZone '+09:00'`. It parses the same value fine without the space, so `timestamp_offsets` drops it for string literals cast to `timestamp_tz` or `timestamp_ltz`, which also covers `to_timestamp_tz()`. All four shapes Snowflake documents now work, and the offset is applied rather than ignored.

**The 500.** `ConversionException` wasn't in the list of duckdb exceptions translated in `_execute`, so anything it raised surfaced as a server error. Clients treat that as retryable: the python connector repeated one failing insert 333 times, the JDBC driver 7. It's now a `ProgrammingError` with `100035 / 22007`, which is what a real account returns for an unrecognised timestamp.

Not covered: a literal in `INSERT INTO t VALUES ('2026-01-01 10:00:00 +09:00')`. There's no cast to hang the rewrite off, and knowing the target column type means reaching for the schema, which the transforms don't do. It fails with the error above rather than a 500 now. Happy to look at it if you have a preference for how, otherwise it seemed better left out of this change.

Rendering the stored offset back is still option 2, untouched here.

Refs #372

### DIFF
--- a/fakesnow/cursor.py
+++ b/fakesnow/cursor.py
@@ -280,6 +280,7 @@ class FakeSnowflakeCursor:
             .transform(transforms.sequence_nextval)
             .transform(transforms.values_columns)
             .transform(transforms.to_date)
+            .transform(transforms.timestamp_offsets)
             .transform(transforms.to_decimal)
             .transform(transforms.to_timestamp)
             .transform(transforms.to_variant)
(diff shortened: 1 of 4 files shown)


tekumara:
thanks for this!

tekumara (approved):

Issue #371, opened 2026-08-10
Batch inserts fail: describeOnly ignored (JDBC 500), CLIENT_STAGE_ARRAY_BINDING_THRESHOLD missing (python qmark)


anjeongkyun:
Batch inserts over array binding don't work. Two clients, seemingly one root cause.

**python, `paramstyle="qmark"`**

```python
conn = snowflake.connector.connect(..., paramstyle="qmark")
cur.execute("create or replace table b (id int)")
cur.executemany("insert into b values (?)", [(1,), (2,), (3,)])
# KeyError: 'CLIENT_STAGE_ARRAY_BINDING_THRESHOLD'
```

The connector reads that session parameter to decide between inline bindings and the stage path. `login_request` returns only `AUTOCOMMIT` and `CLIENT_SESSION_KEEP_ALIVE_HEARTBEAT_FREQUENCY`, so the lookup raises. `paramstyle="pyformat"` works, since the connector expands the statement client-side instead.

**JDBC `executeBatch()`** returns 500, and the driver retries it 7 times. Server log:

```
sql_text='INSERT INTO b VALUES (?)' params=None Unhandled exception
_duckdb.InvalidInputException: Invalid Input Error: Values were not provided for the following prepared statement parameters: 1
```

The request carries `describeOnly=true` with `bindings=None` — the driver asks for metadata before sending the batch, and `query_request` doesn't look at that flag, so the statement runs with nothing bound.

This blocks batching in Hibernate and Spring Data JDBC, and the driver reports `supportsBatchUpdates() == true`, so callers have no way to detect it up front.

Happy to work on this. Two things I'd want your call on first:

- should `describeOnly` return the described rowtype without executing?
- is inline array binding worth supporting, eg: returning `CLIENT_STAGE_ARRAY_BINDING_THRESHOLD` of 0 to keep clients off the stage path?

tekumara:
1. yep we should match snowflake's behaviour and response payload for `describeOnly=true` requests
2. we don't have good internal stage support at the moment so setting  CLIENT_STAGE_ARRAY_BINDING_THRESHOLD=0 to avoid a PUT for large payloads is fine for now

If you're happy to work on these, and they can be reproduced and tested independently, could I suggest a PR for each issue?

anjeongkyun:
Split as suggested, they reproduce and test independently.

[...]

PR #383, opened 2026-08-17
fix: support array bindings, ie: executemany over the server

anjeongkyun:
Part 2 of #371, the python connector half. `describeOnly` is untouched, I'll send that separately.

**The threshold.** Clients read `CLIENT_STAGE_ARRAY_BINDING_THRESHOLD` from the login response to decide whether to bind an array inline or stage it first. It wasn't there, so `executemany` died with `KeyError: 'CLIENT_STAGE_ARRAY_BINDING_THRESHOLD'` before sending anything. Returning 0 keeps the values inline and avoids the PUT, as you suggested.

**The array binding.** With the threshold in place the connector gets further and sends the values, which turned out to be the second half of the problem. An array binding is one list per placeholder rather than one value:

```
{'1': {'type': 'FIXED', 'value': ['1', '2', '3']}, '2': {'type': 'TEXT', 'value': ['one', 'two', 'three']}}
```

`from_binding` assumed a scalar, so `int(value)` got a list and the request 500'd. They're now transposed into rows and executed one at a time, the same tradeoff `cursor.executemany` already makes locally.

The rowcount needed handling because of that: the client sent one statement and expects one answer. Snowflake returns a single `number of rows inserted` for the whole batch, 3 for three rows, so the per-row counts are summed rather than reporting the last one.

Verified end to end against the python connector with `paramstyle="qmark"`. The JDBC driver takes the same path but stops earlier on `describeOnly`, so it stays broken until part 1 lands.

Refs #371

### DIFF
--- a/fakesnow/server.py
+++ b/fakesnow/server.py
@@ -195,7 +195,7 @@ async def query_request(request: Request) -> JSONResponse:
                         {"name": "TIMEZONE", "value": "Etc/UTC"},
                     ],
                     "rowtype": rowtype,
-                    "total": batch_rowcount if batch is not None else cur._rowcount,  # noqa: SLF001
+                    "total": 1 if batch is not None else cur._rowcount,  # noqa: SLF001
                     "queryId": cur.sfqid,
                     "statementTypeId": type_id,
                     "finalDatabaseName": conn.database,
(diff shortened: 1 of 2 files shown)


tekumara:
Thank you! 

ps: see [5d7a2ef](https://github.com/tekumara/fakesnow/pull/383/commits/5d7a2efdb8aa854d0c546688e8116914020f8b78) for a small change I made to match real snowflake.

tekumara (approved):

PR #384, opened 2026-08-17
fix: describe statements without running them for describeOnly requests

### LINKED ISSUE #371: Batch inserts fail: describeOnly ignored (JDBC 500), CLIENT_STAGE_ARRAY_BINDING_THRESHOLD missing (python qmark)

anjeongkyun:
Part 1 of #371. #383 is the other half.

`describeOnly=true` was ignored, so we ran the statement. A describe request carries no bindings, so `insert into example values (?, ?)` arrived with `params=None` and 500'd, which the JDBC driver retried 6 times before giving up.

Now we describe without running. For DML the result columns are the affected row counts, which we already have as SQL, so those get described instead of the statement. For a query the placeholders are replaced with null, which makes it describable without changing the columns it returns.

I checked the response against an account to see what to send back:

| statement | rowtype | arrayBindSupported |
| --- | --- | --- |
| insert | number of rows inserted | true |
| update | number of rows updated, number of multi-joined rows updated | false |
| delete | number of rows deleted | false |
| merge | number of rows inserted, number of rows updated | false |
| create table | status | false |
| select | the query's columns | false |

Snowflake sends `total: 0`, no rows, and json for everything except a query. It also sends `numberOfBinds` and `arrayBindSupported`, which is the field clients read to decide whether they can send a batch as an array binding, and it's only true for inserts. Our existing DML result SQL already produces those exact column names, so the rowtypes line up.

Not covered: statements we can only describe by running, ie: MERGE, SHOW and DDL. MERGE derives its columns from the rows it matched, and SHOW and DDL build their result during execution. They fall through to the old behaviour rather than erroring, so they're no worse than before. Happy to take those on separately if you want them.

With both parts a JDBC batch insert works end to end, `executeBatch` returns `[1, 1, 1]` and the rows land.

Fixes #371

### DIFF
--- a/fakesnow/cursor.py
+++ b/fakesnow/cursor.py
@@ -64,6 +64,15 @@ SQL_COPY_ROWS = Template(
 )
 
 
+# The result columns snowflake reports when asked to describe a DML statement without running it.
+# The counts are placeholders, only the column names and types are used.
+DESCRIBE_RESULT_SQL: dict[type[Expr], str] = {
+    exp.Insert: SQL_INSERTED_ROWS.substitute(count=0),
+    exp.Update: SQL_UPDATED_ROWS.substitute(count=0),
+    exp.Delete: SQL_DELETED_ROWS.substitute(count=0),
+}
+
+
 class FakeSnowflakeCursor:
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


tekumara:
@anjeongkyun looks like there's now a conflict after merging #383

anjeongkyun:
Rebased. The conflict was just my describeOnly block landing next to the `batch_rowcount` from #383, both kept.

Also noticed your fix in 5d7a2ef while rebasing, `total` being the result row count rather than the affected rows. That's consistent with what this PR sends for describeOnly, `total: 0` with an empty rowset, which is what an account returns.

tekumara (approved):
nice work!

PR #386, opened 2026-08-17
fix: typed CTAS with aliased SELECT expression

### LINKED ISSUE #385: CREATE TABLE t (cols) AS SELECT ... AS <alias> emits CAST(x AS <alias> AS <type>) instead of CAST(x AS <type>) AS <alias>
Hey folks! First of all, thank you to the maintainers and contributors, this is a great library. 

## Details

We're looking to use `fakesnow` for a testing framework, to avoid connecting to a real Snowflake instance. One issue we noticed as we were testing internally was with the transpilation of a CTAS statement of this form:

```sql
CREATE TABLE ... (<col_id> <snowflake_type>, ...) AS 
SELECT 
   <val> AS <col_id>
```

This, when transpiled, results in:

```sql
CREATE TABLE ... AS 
SELECT 
  CAST(<val> AS <col_id> AS <duckdb_type>) AS <col_id>
```

which results in a `ProgrammingError: 001003 (42000): Parser Error: syntax error at or near "AS"`. 

I believe it should look like this instead: 

```sql
CREATE TABLE ... AS 
SELECT 
   CAST(<val> AS <duckdb_type>) AS <col_id>
```

## Fix

I would be happy to raise a PR to fix this. [The issue appears to originate here:](https://github.com/tekumara/fakesnow/blob/d6c2071f256503bc2e02b46ac6d76458bed0166f/fakesnow/transforms/transforms.py#L1378) 

```python
cast_expr = exp.Cast(this=select_col, to=create_col_type)
```

I monkeypatched over the `create_table_as` function locally to unwrap any aliases first:

```python
inner = select_col.this if isinstance(select_col, exp.Alias) else select_col
cast_expr = exp.Cast(this=inner, to=create_col_type)
```

and this was sufficient. Please let me know if there are any concerns with this approach. 

## Reproduction

**Reproduced on:**

`fakesnow 0.11.13`, MacOS 26.5.2 (25F84)

<details>

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "fakesnow>=0.11.13",
# ]
# ///
"""Repro: fakesnow mistranslates typed CTAS when the SELECT uses aliases.

Reproduced on:
  fakesnow 0.11.13
  snowflake-connector-python 4.6.0
  duckdb 1.5.5
  sqlglot 30.15.0

Run:
  ./typed_ctas_with_aliases.py
"""

from __future__ import annotations

import fakesnow
import snowflake.connector


CASES = [
    (
        "FAIL — typed column list + SELECT aliases",
        "CREATE OR REPLACE TABLE repro.t1 (id INT, payload VARIANT) "
        "AS SELECT 1 AS id, parse_json('{\"a\":1}') AS payload",
    ),
    (
        "PASS — typed column list, no SELECT aliases",
        "CREATE OR REPLACE TABLE repro.t2 (id INT, payload VARIANT) "
        "AS SELECT 1, parse_json('{\"a\":1}')",
    ),
    (
        "PASS — SELECT aliases, no typed column list",
        "CREATE OR REPLACE TABLE repro.t3 "
        "AS SELECT 1 AS id, parse_json('{\"a\":1}') AS payload",
    ),
    (
        "PASS — create then insert (aliases OK)",
        "CREATE OR REPLACE TABLE repro.t4 (id INT, payload VARIANT)",
    ),
]


def main() -> None:
            account="fakesnow",
            user="test",
            password="test",
            database="TESTDB",
            schema="PUBLIC",
        )
        cur = conn.cursor()
        cur.execute("CREATE SCHEMA IF NOT EXISTS repro")

        for label, sql in CASES:
            try:
                cur.execute(sql)
                print(f"PASS  {label}")
            except Exception as e:
                print(f"FAIL  {label}")
                print(f"      {type(e).__name__}: {e}")

        try:
            cur.execute(
                "INSERT INTO repro.t4 SELECT 1 AS id, parse_json('{\"a\":1}') AS payload"
            )
            print("PASS  PASS — create then insert (aliases OK) [insert]")
        except Exception as e:
            print(f"FAIL  insert into typed table: {e}")


if __name__ == "__main__":
    main()
```

Local output:

```
./typed_ctas_with_aliases.py                                                                                                                                      
FAIL  FAIL — typed column list + SELECT aliases
      ProgrammingError: 001003 (42000): Parser Error: syntax error at or near "AS"

LINE 1: CREATE OR REPLACE TABLE REPRO.T1 AS SELECT CAST(1 AS ID AS BIGINT) AS ID, CAST(JSON('{"a":1}') AS PAYLOAD AS JSON...
                                                                ^
PASS  PASS — typed column list, no SELECT aliases
PASS  PASS — SELECT aliases, no typed column list
PASS  PASS — create then insert (aliases OK)
PASS  PASS — create then insert (aliases OK) [insert]
```

</summary>

ishaan-upadhyay:
## Summary

Typed CTAS statements with aliased SELECT expressions fail even though they are valid Snowflake SQL

Fixes #385 

## Details

`create_table_as` wraps each SELECT expression in `CAST(... AS <declared_type>)` then re-aliases it to the declared column name. When the SELECT expression is already an `Alias` (`1 AS id`), that becomes invalid DuckDB:

```sql
-- input
CREATE TABLE t (id INT, payload VARIANT) AS
SELECT 1 AS id, parse_json('{"a":1}') AS payload

-- before (broken)
CAST(1 AS ID AS BIGINT) AS ID, CAST(JSON(...) AS PAYLOAD AS JSON) AS PAYLOAD

-- after
CAST(1 AS BIGINT) AS ID, CAST(JSON(...) AS JSON) AS PAYLOAD
```

Fix: unwrap `exp.Alias` before casting so we cast the value, then alias.

## Test plan

- [x] Extended `test_create_table_as` with the previously-failing typed-cols + SELECT-aliases form
- [x] `pytest tests/test_fakes.py::test_create_table_as` passes locally

Made with [Cursor](https://cursor.com)

### DIFF
--- a/fakesnow/transforms/transforms.py
+++ b/fakesnow/transforms/transforms.py
@@ -1375,8 +1375,6 @@ def create_table_as(expression: Expr, duck_conn: DuckDBPyConnection) -> Expr:
             create_col_type = col_def.kind
             select_col = select_query.expressions[i]
 
-            # Unwrap an existing Alias so we cast the value, not `val AS alias`.
-            # Otherwise DuckDB receives invalid SQL like `CAST(1 AS ID AS BIGINT)`.
             inner = select_col.this if isinstance(select_col, exp.Alias) else select_col
             cast_expr = exp.Cast(this=inner, to=create_col_type)
             aliased_expr = exp.Alias(this=cast_expr, alias=create_col_id)
--- a/fakesnow/transforms/transforms.py
+++ b/fakesnow/transforms/transforms.py
@@ -1375,7 +1375,10 @@ def create_table_as(expression: Expr, duck_conn: DuckDBPyConnection) -> Expr:
             create_col_type = col_def.kind
             select_col = select_query.expressions[i]
 
-            cast_expr = exp.Cast(this=select_col, to=create_col_type)
+            # Unwrap an existing Alias so we cast the value, not `val AS alias`.
+            # Otherwise DuckDB receives invalid SQL like `CAST(1 AS ID AS BIGINT)`.
+            inner = select_col.this if isinstance(select_col, exp.Alias) else select_col
+            cast_expr = exp.Cast(this=inner, to=create_col_type)
             aliased_expr = exp.Alias(this=cast_expr, alias=create_col_id)
 
             new_expressions.append(aliased_expr)
--- a/tests/test_fakes.py
+++ b/tests/test_fakes.py
@@ -187,6 +187,14 @@ def test_create_table_as(dcur: snowflake.connector.cursor.SnowflakeCursor) -> No
     dcur.execute("select * from t1")
     assert dcur.fetchall() == [{"ID": "1"}]
 
+    # Typed column list + SELECT aliases previously emitted
+    # `CAST(1 AS ID AS BIGINT) AS ID` which DuckDB rejects.
+    dcur.execute(
+        "create or replace table t1(id int, payload variant) as select 1 as id, parse_json('{\"a\":1}') as payload"
+    )
+    dcur.execute("select id, payload from t1")
+    assert dcur.fetchall() == [{"ID": 1, "PAYLOAD": '{"a":1}'}]
+
 
 def test_dateadd_date_cast(dcur: snowflake.connector.DictCursor):
     q = """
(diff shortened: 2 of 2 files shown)


tekumara (approved):
thank you for your first contribution!

PR #388, opened 2026-08-18
fix: support bind params and a values list source in MERGE

### LINKED ISSUE #387: MERGE statement fails with HTTP 500 when executed via JDBC PreparedStatement with any bind parameter
## Summary

When executing a `MERGE` statement through the Snowflake JDBC driver (`snowflake-jdbc`) against fakesnow, the request fails with HTTP 500 ("Unhandled exception") whenever the `PreparedStatement` has **any bind parameter**, regardless of the exact MERGE syntax used. This happens with `ghcr.io/tekumara/fakesnow:0.11.12`.

## Reproduction pattern (isolated via plain JDBC, no Spring/ORM involved)

Tested directly against the fakesnow container with the real `snowflake-jdbc` driver:

- ✅ A plain bound `INSERT ... VALUES (?, ?, ...)` via `PreparedStatement` works correctly.
- ✅ A `MERGE` statement with only **literal** values (no bind parameters) works correctly.
- ❌ A `MERGE` statement with **any** bind parameter (`?`) fails with HTTP 500, regardless of the specific MERGE syntax variant:
  - `MERGE ... USING (SELECT ? AS col ...) ...`
  - `MERGE ... USING (VALUES (?, ...)) ...`
  - Variants using explicit `CAST(? AS ...)` on the bound parameter

The error surfaces as an HTTP 500 with body similar to:
```json
{"code":"261000","message":"...Unhandled exception..."}
```

## Impact

This makes it impossible to use `MERGE ... WHEN MATCHED ... WHEN NOT MATCHED ...` (a common upsert pattern) via JDBC `PreparedStatement` with bind parameters against fakesnow, even though the equivalent plain `INSERT`/`UPDATE` with bind parameters works fine. The only workaround we found was rewriting the MERGE as a sequence of simple statements (`UPDATE` → check-exists → `INSERT`).

## Environment

- Image: `ghcr.io/tekumara/fakesnow:0.11.12`
- Driver: `net.snowflake:snowflake-jdbc:3.19.0`
- Client: plain JDBC (`java.sql.PreparedStatement`), also reproduced via Spring's `JdbcTemplate`

Happy to provide a minimal standalone Java reproduction (bypassing Spring/Gradle entirely) if useful — let me know and I'll attach it.

anjeongkyun:
Two independent causes behind #387, both giving a 500.

**Params bound to every part of a split statement.** A MERGE becomes four statements, and only the one building `merge_candidates` keeps the placeholders. The same params were bound to all four, so the other three failed with `Parameter argument/count mismatch, identifiers of the excess parameters: 1, 2`. They're now bound to the parts that have placeholders.

This is why the issue reported that any bind parameter breaks a MERGE, while a bound INSERT is fine: an INSERT isn't split.

**A values list as the source.** `USING (VALUES (1, 'a')) AS s (id, name)` parses to `exp.Values`, not `exp.Subquery`, and only a subquery was checked for the alias, so the name was read from `.this`, wasn't an `Identifier`, and the assert blew up.

Worth noting this one has nothing to do with binds, it fails with literals too:

```sql
MERGE INTO t tgt USING (VALUES (1, 'x')) src (id, name) ON tgt.id = src.id
WHEN MATCHED THEN UPDATE SET tgt.name = src.name
```

I checked all three shapes from the issue against an account and Snowflake accepts them, including the values list.

Reproduces without the server, on `paramstyle="qmark"`, so the tests are in `test_merge.py` rather than `test_server.py`. Also confirmed through the JDBC driver, which is where it was found: all three variants pass now, previously two retried into a 500 and one hit the assert.

Fixes #387

### DIFF
--- a/fakesnow/cursor.py
+++ b/fakesnow/cursor.py
@@ -177,9 +177,11 @@ class FakeSnowflakeCursor:
             expression = parse_one(command, read="snowflake")
             self.check_db_and_schema(expression)
 
-            for exp in self._transform_explode(expression):
-                transformed = self._transform(exp, params)
-                self._execute(transformed, params)
+            for statement in self._transform_explode(expression):
+                transformed = self._transform(statement, params)
+                # a statement split into several, eg: MERGE, keeps the placeholders in whichever
+                # part reads the source rows. duckdb rejects params for the parts without any.
+                self._execute(transformed, params if transformed.find(exp.Placeholder) else None)
 
             if not kwargs.get("server") and (put_stage_data := transformed.args.get("put_stage_data")):  # pyright: ignore[reportPossiblyUnboundVariable]
                 self._put_files(put_stage_data)
--- a/fakesnow/transforms/merge.py
+++ b/fakesnow/transforms/merge.py
@@ -22,7 +22,13 @@ def _create_merge_candidates(merge_expr: exp.Merge) -> Expr:
 
     source = merge_expr.args.get("using")
     assert isinstance(source, Expr)
-    source_id = (alias := source.args.get("alias")) and alias.this if isinstance(source, exp.Subquery) else source.this
+    # a source is either a table, or a subquery or values list that carries its name in an alias,
+    # eg: USING (VALUES (1, 'a')) AS s (id, name)
+    source_id = (
+        (alias := source.args.get("alias")) and alias.this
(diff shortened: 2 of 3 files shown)


tekumara (approved):
ty!

PR #390, opened 2026-08-18
feat: support structured types (by treating as JSON without their element types)

### LINKED ISSUE #389: Structured types (`ARRAY(T)`, `OBJECT(k T)`) fail with "Expected a constant as type modifier"
# Details

Snowflake supports both:
- [semi-structured types](https://docs.snowflake.com/en/sql-reference/data-types-semistructured) (`ARRAY`, `OBJECT`, `VARIANT`) 
- [structured types](https://docs.snowflake.com/en/sql-reference/data-types-structured) (`ARRAY(VARCHAR)`, `OBJECT(a INT)`, `MAP(VARCHAR, INT)`). 

`fakesnow` handles the semi-structured forms but rejects the structured ones at the DuckDB parse step, as it retains nested type information:

```python
import fakesnow, snowflake.connector

        account="fakesnow", user="test", password="test",
        database="TESTDB", schema="PUBLIC",
    )
    cur = conn.cursor()
    cur.execute("CREATE TABLE t1 (c ARRAY)")           # ok

    # This is transformed to "CREATE TABLE T2 (C JSON(TEXT))"
    cur.execute("CREATE TABLE t2 (c ARRAY(VARCHAR))")  # ProgrammingError

>>> snowflake.connector.errors.ProgrammingError: 001003 (42000):
>>> Parser Error: Expected a constant as type modifier
```

This also affects nested appearances of structured types (e.g. `MAP(VARCHAR, INT)` works with `fakesnow`, but `MAP(VARCHAR, ARRAY(INT))` does not). 

# Possible Fix 

[`semi_structured_types()`](https://github.com/tekumara/fakesnow/blob/d6c2071f256503bc2e02b46ac6d76458bed0166f/fakesnow/transforms/transforms.py#L1269) maps these types onto DuckDB's `JSON` by copying the `DataType` node and reassigning only `this`:

```python
new = expression.copy()
new.args["this"] = exp.DataType.Type.JSON
return new
```

Instead, we could create a fresh `JSON` node and discard any nested information (such as the element types): 

```python
if isinstance(expression, exp.DataType) and expression.this in (
    exp.DataType.Type.ARRAY,
    exp.DataType.Type.OBJECT,
    exp.DataType.Type.VARIANT,
):
    return exp.DataType(this=exp.DataType.Type.JSON, nested=False)
```

I alternatively considered [using DuckDB native types](https://duckdb.org/docs/lts/sql/data_types/list) (e.g. converting `ARRAY(VARCHAR)` -> `VARCHAR[]`), but I saw this was a longer standing issue (#26, #27) and wasn't sure if a larger-scale refactor would be ideal. 

Would be happy to submit a PR if this seems reasonable.

# Reproducing (in detail)

I also monkey-patched in the proposed fix to demonstrate viability.

**Reproduced on:**

`fakesnow 0.11.13`, MacOS 26.5.2 (25F84)

<details>

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "fakesnow>=0.11.13",
# ]
# ///
"""Repro: fakesnow rejects Snowflake structured types such as ARRAY(VARCHAR).

Expected output:

    current
       ok    ARRAY
       FAIL  ARRAY(VARCHAR)   ... Expected a constant as type modifier
       ok    OBJECT
       FAIL  OBJECT(a INT)    ... syntax error at or near "BIGINT"

    with the proposed fix
       ok    ARRAY
       ok    ARRAY(VARCHAR)
       ok    OBJECT
       ok    OBJECT(a INT)
"""

import fakesnow
import fakesnow.transforms as transforms
import snowflake.connector
from sqlglot import exp

TYPES = ["ARRAY", "ARRAY(VARCHAR)", "OBJECT", "OBJECT(a INT)"]


def patched(expression: exp.Expression) -> exp.Expression:
    """semi_structured_types() returning a bare JSON node, without the element types."""
    if isinstance(expression, exp.DataType) and expression.this in (
        exp.DataType.Type.ARRAY,
        exp.DataType.Type.OBJECT,
        exp.DataType.Type.VARIANT,
    ):
        return exp.DataType(this=exp.DataType.Type.JSON, nested=False)
    return expression


def create_all(cur, tag: str) -> None:
    for i, dtype in enumerate(TYPES):
        try:
            cur.execute(f"CREATE TABLE {tag}{i} (c {dtype})")
            print(f"   ok    {dtype}")
        except snowflake.connector.errors.ProgrammingError as e:
            print(f"   FAIL  {dtype:<16} {str(e).splitlines()[0]}")


    cur = snowflake.connector.connect(
        account="fakesnow", user="test", password="test",
        database="TESTDB", schema="PUBLIC",
    ).cursor()

    print("current")
    create_all(cur, "before")

    # cursor.py looks this up on the package at call time, so replacing the
    # attribute is enough to swap the transform.
    transforms.semi_structured_types = patched

    print("\nwith the proposed fix")
    create_all(cur, "after")

```

Local output:

```
./parameterized_array_types.py                                                                                                     
current
   ok    ARRAY
   FAIL  ARRAY(VARCHAR)   001003 (42000): Parser Error: Expected a constant as type modifier
   ok    OBJECT
   FAIL  OBJECT(a INT)    001003 (42000): Parser Error: syntax error at or near "BIGINT"

with the proposed fix
   ok    ARRAY
   ok    ARRAY(VARCHAR)
   ok    OBJECT
   ok    OBJECT(a INT)
```

</summary>

ishaan-upadhyay:
## Why

Closes #389   

Snowflake supports structured type declarations like `ARRAY(VARCHAR)` or `OBJECT(a INT)`. `fakesnow` converts these to `JSON` types, but retains the nested type information, causing DuckDB to emit `Parser Error: Expected a constant as type modifier`. 

## How

Convert Snowflake structured types to plain DuckDB `JSON` types while preserving their nullability constraints. This allows structured values to use the same JSON-backed representation as existing semi-structured types.

## Details 

`semi_structured_types()` copies the `DataType` node and reassigns only `this`, so a Snowflake structured type keeps its `expressions`:

There are are a few `arg_types` on a `DataType` node: 

```python
arg_types = {
    "this": True,        # required
    "expressions": False,
    "nested": False,
    "values": False,
    "kind": False,
    "nullable": False,
    "collate": False,
}
```

Returns a fresh `JSON` node instead, dropping the optional `arg_types` to prevent them from flowing through as type modifiers.

---

I found Snowflake's naming a bit confusing and wasn't sure if that needs disambiguation here: 

> [The Snowflake structured types are ARRAY, OBJECT, and MAP.](https://docs.snowflake.com/en/sql-reference/data-types-structured)

(from Snowflake docs)

 However, the types share names with semi-structured types, leading to weird test names like `test_semi_structured_types_structured()`, if that should be cleaned up, please let me know

## Testing

Added `test_semi_structured_types_structured`, covering `ARRAY(T)`, `OBJECT(k T)` and a structured type nested inside a `MAP`.

### DIFF
--- a/tests/test_transforms.py
+++ b/tests/test_transforms.py
@@ -722,6 +722,19 @@ def test_semi_structured_types_structured() -> None:
         == "CREATE TABLE table1 (name JSON NOT NULL)"
     )
 
+    assert (
+        sqlglot.parse_one("SELECT col::ARRAY(VARCHAR)", read="snowflake")
+        .transform(semi_structured_types)
+        .sql(dialect="duckdb")
+        == "SELECT CAST(col AS JSON)"
+    )
+    assert (
+        sqlglot.parse_one("SELECT col::OBJECT(a INT)", read="snowflake")
+        .transform(semi_structured_types)
+        .sql(dialect="duckdb")
+        == "SELECT CAST(col AS JSON)"
+    )
+
 
 def test_show_tables_etc() -> None:
     def _show_tables_etc(e: Expr) -> Expr:
--- a/tests/test_transforms.py
+++ b/tests/test_transforms.py
@@ -701,6 +701,7 @@ def test_semi_structured_types_structured() -> None:
         "CREATE TABLE table1 (name ARRAY(ARRAY(INT)))",
         "CREATE TABLE table1 (name OBJECT(a INT))",
         "CREATE TABLE table1 (name OBJECT(a INT, b VARCHAR))",
+        "CREATE TABLE table1 (name OBJECT(a INT NOT NULL))",
     ]:
         assert (
             sqlglot.parse_one(sql, read="snowflake").transform(semi_structured_types).sql(dialect="duckdb")
--- a/tests/test_transforms.py
+++ b/tests/test_transforms.py
@@ -701,7 +701,6 @@ def test_semi_structured_types_structured() -> None:
         "CREATE TABLE table1 (name ARRAY(ARRAY(INT)))",
         "CREATE TABLE table1 (name OBJECT(a INT))",
         "CREATE TABLE table1 (name OBJECT(a INT, b VARCHAR))",
-        "CREATE TABLE table1 (name OBJECT(a INT NOT NULL))",
     ]:
         assert (
             sqlglot.parse_one(sql, read="snowflake").transform(semi_structured_types).sql(dialect="duckdb")
--- a/tests/test_transforms.py
+++ b/tests/test_transforms.py
@@ -695,6 +695,27 @@ def test_semi_structured_types() -> None:
     )
 
 
+def test_semi_structured_types_structured() -> None:
+    for sql in [
+        "CREATE TABLE table1 (name ARRAY(VARCHAR))",
+        "CREATE TABLE table1 (name ARRAY(ARRAY(INT)))",
+        "CREATE TABLE table1 (name OBJECT(a INT))",
+        "CREATE TABLE table1 (name OBJECT(a INT, b VARCHAR))",
+    ]:
+        assert (
+            sqlglot.parse_one(sql, read="snowflake").transform(semi_structured_types).sql(dialect="duckdb")
+            == "CREATE TABLE table1 (name JSON)"
+        )
+
+    # structured types nested inside a type that isn't itself converted
+    assert (
+        sqlglot.parse_one("CREATE TABLE table1 (name MAP(VARCHAR, ARRAY(INT)))", read="snowflake")
+        .transform(semi_structured_types)
+        .sql(dialect="duckdb")
+        == "CREATE TABLE table1 (name MAP(TEXT, JSON))"
+    )
+
+
 def test_show_tables_etc() -> None:
     def _show_tables_etc(e: Expr) -> Expr:
         return show_tables_etc(e, None, None)
--- a/tests/test_transforms.py
+++ b/tests/test_transforms.py
@@ -715,6 +715,13 @@ def test_semi_structured_types_structured() -> None:
         == "CREATE TABLE table1 (name MAP(TEXT, JSON))"
     )
 
+    assert (
+        sqlglot.parse_one("CREATE TABLE table1 (name OBJECT(a INT NOT NULL) NOT NULL)", read="snowflake")
+        .transform(semi_structured_types)
+        .sql(dialect="duckdb")
+        == "CREATE TABLE table1 (name JSON NOT NULL)"
+    )
+
 
 def test_show_tables_etc() -> None:
     def _show_tables_etc(e: Expr) -> Expr:
--- a/fakesnow/transforms/transforms.py
+++ b/fakesnow/transforms/transforms.py
@@ -1269,6 +1269,10 @@ def try_parse_json(expression: Expr) -> Expr:
 def semi_structured_types(expression: Expr) -> Expr:
     """Convert OBJECT, ARRAY, and VARIANT types to duckdb compatible types.
 
+    Structured types, eg: ARRAY(VARCHAR) or OBJECT(a INT), become JSON too. Their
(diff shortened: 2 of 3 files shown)


tests/test_transforms.py
```diff
@@ -695,6 +695,28 @@ def test_semi_structured_types() -> None:
     )
 
 
+def test_semi_structured_types_structured() -> None:
```
tekumara, line 698: can we add not a transformation test but one that verifies reading from the field too?

tests/test_transforms.py
```diff
@@ -695,6 +695,28 @@ def test_semi_structured_types() -> None:
...
+    for sql in [
+        "CREATE TABLE table1 (name ARRAY(VARCHAR))",
+        "CREATE TABLE table1 (name ARRAY(ARRAY(INT)))",
+        "CREATE TABLE table1 (name OBJECT(a INT))",
+        "CREATE TABLE table1 (name OBJECT(a INT, b VARCHAR))",
+        "CREATE TABLE table1 (name OBJECT(a INT NOT NULL))",
```
tekumara: this fails in real snowflake:

```
093209 (23001): SQL compilation error: DDL operation failed because it would result in a
       non-nullable structured type field 'NAME.a' contained within a nullable object.
```

the correct form is `CREATE TABLE table1 (name OBJECT(a INT NOT NULL) NOT NULL)` or we can remove it

ishaan-upadhyay: Ah I see, I was basing it on [this part of the Snowflake docs specifying the declaration of such a type](https://docs.snowflake.com/en/sql-reference/data-types-structured#specifying-a-structured-object-type), but did not verify in real Snowflake:

```sql
OBJECT(
  [
    <key> <value_type> [ NOT NULL ]
    [ , <key> <value_type> [ NOT NULL ] ]
    [ , ... ]
  ]
)
```

Added a test for the correct form

tests/test_transforms.py
ishaan-upadhyay: Yep, added a couple in `test_semis.py`. Just to double-check w/ you...

These queries from the tests will fail in real Snowflake because they are inserting semi-structured data into structured columns:

```sql
# 001010 (42846) incompatible types
insert into structured(names, attrs)
select ['A', 'B'], object_construct('a', 1);

insert into structured_nn(attrs)
select object_construct('a', 1);
```

The correct Snowflake syntax would be with casts ([or use of functions `ARRAY_CONSTRUCT_STRUCTURED` & `OBJECT_CONSTRUCT_KEEP_NULL_STRUCTURED`](https://docs.snowflake.com/en/sql-reference/data-types-structured#using-sql-functions-to-construct-structured-array-and-object-values))

```sql
insert into structured(names, attrs)
select
    ['A', 'B']::array(varchar),
    object_construct('a', 1)::object(a int);

insert into structured_nn(attrs)
select object_construct('a', 1)::object(a int not null);
```

But given in `fakesnow` we are using `JSON` under the hood, I imagine this might fall under the caveat from the `README.md`: 

> fakesnow supports a more liberal SQL dialect than actual Snowflake. This means some queries that work with fakesnow might not work with a real Snowflake instance.

tekumara: yes well spotted - its ok if fakesnow is more liberal, but we don't want to bake in tests that would fail if run on a real snowflake instance, ie: the tests should be an accurate reflection of real snowflake behaviour even if fakesnow isn't.

ishaan-upadhyay: Yes, that makes sense.

 I've updated the tests to assert on the form with casting on `INSERT`. I also added casting on the `SELECT` to account for the `JSON` storage under the hood. 

(The casting on `SELECT` would be redundant in real Snowflake, however I have added it here so that the tests would work as is in both places.)

tekumara (approved):
nice work ty!

PR #392, opened 2026-08-23
fix: describe DDL and MERGE without running them

anjeongkyun:
### Why?

Preparing DDL or MERGE statements currently executes them, causing side effects before the caller requests execution. For example, describing a DROP can remove the table, while describing a MERGE can modify rows.

This differs from Snowflake, which returns result metadata without executing the statement.

### How?

Return Snowflake-compatible result metadata for DDL and MERGE statements without executing them.

### Details

`describeOnly` fell through to executing anything that wasn't a query or a plain INSERT/UPDATE/DELETE, so preparing a statement had side effects. The clearest way to see it is that the test I added fails on main like this:

```
assert [m.name for m in cur.describe("drop table example")] == ["status"]
>  assert [m.name for m in cur.describe("alter table example add column w int")] == ["status"]
E  ProgrammingError: 002003 (42S02): Catalog Error: Table with name EXAMPLE does not exist!
```

Describing the DROP dropped the table. Describing a MERGE moves rows the same way. `connection.prepareStatement(...)` is enough to trigger it, no execute needed.

An account doesn't run these. I checked with `cursor.describe` and then looked for the objects:

| statement | rowtype | created/changed anything |
| --- | --- | --- |
| CREATE TABLE | status | no |
| DROP TABLE | status | no |
| ALTER TABLE | status | no |
| CREATE VIEW | status | no |
| TRUNCATE | status | no |

So DDL is a single `status` column. MERGE reports a count per operation it actually performs, which the WHEN clauses already determine:

| merge | rowtype |
| --- | --- |
| matched update + not matched insert | number of rows inserted, number of rows updated |
| matched delete only | number of rows deleted |
| not matched insert only | number of rows inserted |

That's the same rule `_counts` uses to build the result, so I pulled its WHEN classification out into `operations()` and reused it rather than duplicating the logic.

SHOW still describes by running. It only reads, so there's no side effect, and its columns are built during execution. Happy to take that on separately if you want it.

Stacked on #391, which the delete-only merge case needs.

### DIFF
--- a/fakesnow/cursor.py
+++ b/fakesnow/cursor.py
@@ -33,6 +33,7 @@ from fakesnow.copy_into import copy_into
 from fakesnow.params import MutableParams
 from fakesnow.rowtype import describe_as_result_metadata
 from fakesnow.transforms import stage
+from fakesnow.transforms.merge import operations as merge_operations
 
 if TYPE_CHECKING:
     # don't require pandas at import time
@@ -63,6 +64,11 @@ SQL_COPY_ROWS = Template(
     "NULL as first_error_column_name"
 )
 
+
+# Statements that report a status rather than a result set, so describing them without running
+# them is the same as describing the status. Verified against an account: CREATE, DROP, ALTER and
+# TRUNCATE all describe as a single "status" column.
+DDL_EXPRESSIONS = (exp.Create, exp.Drop, exp.Alter, exp.TruncateTable)
 
 # The result columns snowflake reports when asked to describe a DML statement without running it.
 # The counts are placeholders, only the column names and types are used.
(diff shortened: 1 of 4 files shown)


tekumara:
thanks @anjeongkyun - it looks like there's a conflict now that I merged #391

anjeongkyun:
Rebased onto main. The conflict was the `operations()` extraction landing on top of the inline classification #391 touched, both kept.

tekumara (approved):

tekumara:
Thanks @anjeongkyun for the contribution! I pushed a small follow-up refactor in d93ce8d to simplify the DDL result mapping and consolidate the describe-only coverage. The behaviour remains unchanged.

PR #395, opened 2026-08-26
feat: support * as an argument to OBJECT_CONSTRUCT

### LINKED ISSUE #394: Support `*` + aliased `*` as an argument to `OBJECT_CONSTRUCT()` and related functions
# Details

[`OBJECT_CONSTRUCT()`](https://docs.snowflake.com/en/sql-reference/functions/object_construct) can take a wildcard in place of key/value pairs, optionally qualified with a table alias.

From Snowflake docs:
>When invoked with an asterisk (wildcard), the OBJECT value is constructed from the specified data using the attribute names as keys and the associated values as values. See the examples below.
>
> When you pass a wildcard to the function, you can qualify the wildcard with the name or alias for the table. For example, to pass in all of the columns from the table named mytable, specify the following: `(mytable.*)`

```sql
select object_construct_keep_null(*) as failing_row
from analytics.orders
where order_total < 0
```

Under fakesnow the wildcard forms fail, differently depending on the variant:

```
object_construct(*)            Catalog Error: Scalar Function with name star_map does not exist!
object_construct(t.*)          Catalog Error: Scalar Function with name star_map does not exist!
object_construct_keep_null(*)  AssertionError (empty message)
```

### Root cause

The below is referring to the [`object_construct()` function](https://github.com/tekumara/fakesnow/blob/0649a444cd3ad3eae9fd9f39decffa23d6dead1a/fakesnow/transforms/transforms.py#L848) in `transforms.py`.

`OBJECT_CONSTRUCT(*)` parses to `exp.StarMap`, which `object_construct()` does not match, so it reaches DuckDB as `STAR_MAP(*)` [from the  `else` here](https://github.com/tekumara/fakesnow/blob/0649a444cd3ad3eae9fd9f39decffa23d6dead1a/fakesnow/transforms/transforms.py#L869).

`OBJECT_CONSTRUCT_KEEP_NULL(*)` parses to `exp.JSONObject` holding a `Star` and [trips `assert isinstance(kv, exp.JSONKeyValue)`](https://github.com/tekumara/fakesnow/blob/0649a444cd3ad3eae9fd9f39decffa23d6dead1a/fakesnow/transforms/transforms.py#L866-L867). As this is an `assert`, the message is empty and carries no SQL context.

Based on my reading of the `_FS_OBJECT_CONSTRUCT` macro, it would not be correct to modify it as it is intended to explicitly handle the case where keys and values are known (which they are not here).

# Suggested fix

In the case where the wildcard is qualified, we can reduce to TO_JSON.  While this behaviour is not directly [documented in DuckDB's Creating JSON page](https://duckdb.org/docs/lts/data/json/creating_json), this does work.

```
object_construct_keep_null(t.*)  ->  TO_JSON(t) 
```

For unqualified, we can use `COLUMNS(*)` to discover all the columns which are being referenced, then use `*COLUMNS(*)` to spread them into arguments for `STRUCT_PACK()`

```
object_construct_keep_null(*)    ->  TO_JSON(STRUCT_PACK(*COLUMNS(*)))
```

As for `object_construct()`, we can rely on the [RFC-7396 specification](https://datatracker.ietf.org/doc/html/rfc7396) of `JSON_MERGE_PATCH()` semantics, which DuckDB implements. Specifically, we can rely on: 

>Null values in the merge patch are given special meaning to indicate the removal of existing values in the target.

Resulting in (where `{}` is a dummy patch target, only used for filtering out nulls):

```
object_construct(*)              ->  JSON_MERGE_PATCH('{}', TO_JSON(STRUCT_PACK(*COLUMNS(*))))
object_construct(t.*)            ->  JSON_MERGE_PATCH('{}', TO_JSON(t))
```

I have a rough implementation of this in the script for reproduction + output section.  Test suite with this change: **378 passed, 1 xfailed**, identical to unpatched `0.11.15`.

Happy to submit a PR with tests alongside the existing `OBJECT_CONSTRUCT` ones.

### Snowflake parity

Run on Snowflake over `values (1, 'x'), (2, null) as v(a, b)`, versus patched fakesnow:

| | Snowflake | patched fakesnow |
| --- | --- | --- |
| `object_construct(*)` | `{"A":1,"B":"x"}`, `{"A":2}` | same |
| `object_construct_keep_null(*)` | `{"A":1,"B":"x"}`, `{"A":2,"B":null}` | same |
| `object_construct(t.*)` | `{"A":1,"B":"x"}`, `{"A":2}` | same |
| `keep_null(t.*)`, joined to a second table | only `t`'s columns | same |
| `keep_null(*)`, join sharing a column name | `Duplicate field key 'A'` | `Duplicate struct entry name "A"` |

# Future work

1. In DuckDB 2.0, we can replace `JSON_MERGE_PATCH` with `JSON_STRIP_NULLS` for the base `OBJECT_CONSTRUCT` function ([link](https://duckdb.org/2026/08/18/reconciling-json#json_strip_nulls-recursively-drop-null-valued-keys)). 
2. `object_construct_keep_null(alias.*)` is untestable today (an expected fail to flip later) due to SQLGlot parsing issues. [See upstream issue here](https://github.com/tobymao/sqlglot/issues/8262).
3. Supporting `* EXCLUDE (cols..)` or `* ILIKE <pattern>` syntax within `OBJECT_CONSTRUCT`. I'll file an additional issue for this one, but probably won't be working on it anytime soon. 
    a. In the unqualified `*` case, this reduces to to supplying extra arguments to `COLUMNS(*)`, [as documented here](https://duckdb.org/docs/current/sql/query_syntax/select). 

# Reproducing

I monkey-patched in the proposed fix to demonstrate viability.

**Reproduced on:** `fakesnow 0.11.15`, MacOS 26.5.2

<details>

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "fakesnow>=0.11.15",
# ]
# ///
"""Repro: OBJECT_CONSTRUCT[_KEEP_NULL] cannot take a star argument.

Expected output:

    current
       FAIL  object_construct(*)                ... Catalog Error: Scalar Function with name star_map does not exist!
       FAIL  object_construct_keep_null(*)      ... AssertionError (empty message)
       FAIL  object_construct(t.*)              ... Catalog Error: Scalar Function with name star_map does not exist!

    with the proposed fix
       ok    object_construct(*)                -> ['{"A":1,"B":"x"}', '{"A":2}']
       ok    object_construct_keep_null(*)      -> ['{"A":1,"B":"x"}', '{"A":2,"B":null}']
       ok    object_construct(t.*)              -> ['{"A":1,"B":"x"}', '{"A":2}']

    Snowflake returns the same keys, casing and null handling (pretty-printed,
    as it does for the key/value form too).
"""

import fakesnow
import fakesnow.transforms as transforms
import snowflake.connector
import sqlglot
from sqlglot import Expr, exp

# ---------------------------------------------------------------- proposed fix

original = transforms.object_construct

# A star can carry a column filter, which this lowering does not implement.
_STAR_FILTERS = ("ilike", "except_", "replace", "rename")


def _star_arg(expression: Expr) -> Expr | None:
    """The star argument, when it is one we can lower.

    A filtered star (`* ILIKE 'col1%'`) selects a subset of the columns, so it
    is left alone to keep failing rather than serialising every column.
    """
    if not expression.is_star:
        return None
    star = expression.this if isinstance(expression, exp.Column) else expression
    if any(star.args.get(name) for name in _STAR_FILTERS):
        return None
    return expression


def _star_source(star: Expr) -> Expr:
    """The duckdb expression naming the row a star refers to."""
    table = star.args.get("table") if isinstance(star, exp.Column) else None
    if table is not None:
        # OBJECT_CONSTRUCT(t.*) names one source, which TO_JSON takes directly.
        return exp.Column(this=table.copy())
    # A bare star is every column in scope; STRUCT_PACK collects them into one
    # row value without needing to know the column names here.
    return exp.Anonymous(this="STRUCT_PACK", expressions=[exp.var("*COLUMNS(*)")])


def _star_object(star: Expr, keep_nulls: bool) -> Expr:
    obj = exp.Anonymous(this="TO_JSON", expressions=[_star_source(star)])
    if keep_nulls:
        return obj
    # OBJECT_CONSTRUCT omits null values and TO_JSON keeps them. JSON merge
    # patch deletes members whose patch value is null, dropping exactly those.
    return exp.Anonymous(
        this="JSON_MERGE_PATCH",
        expressions=[exp.Literal(this="{}", is_string=True), obj],
    )


def patched(expression: Expr) -> Expr:
    """object_construct() with a branch for the star forms."""
    if isinstance(expression, exp.StarMap):
        star = _star_arg(expression.this)
        if star is not None:
            return _star_object(star, keep_nulls=False)

    if isinstance(expression, exp.JSONObject) and len(expression.expressions) == 1:
        star = _star_arg(expression.expressions[0])
        if star is not None:
            return _star_object(star, keep_nulls=True)

    return original(expression)


# --------------------------------------------------------------------- harness

# OBJECT_CONSTRUCT_KEEP_NULL(t.*) is omitted: sqlglot cannot parse it yet, so it
# never reaches this transform. The other three do.
SELECTS = [
    "object_construct(*)",
    "object_construct_keep_null(*)",
    "object_construct(t.*)",
]


def run_all(cur) -> None:
    for select in SELECTS:
        try:
            cur.execute(f"SELECT {select} FROM tbl AS t")
            print(f"   ok    {select:<33} -> {[r[0] for r in cur.fetchall()]}")
        except snowflake.connector.errors.ProgrammingError as e:
            print(f"   FAIL  {select:<33} ... {str(e).splitlines()[0].split(': ', 1)[-1]}")
        except AssertionError:
            print(f"   FAIL  {select:<33} ... AssertionError (empty message)")


    cur = snowflake.connector.connect(
        account="fakesnow",
        user="test",
        password="test",
        database="TESTDB",
        schema="PUBLIC",
    ).cursor()

    cur.execute("CREATE TABLE tbl (a INT, b VARCHAR)")
    cur.execute("INSERT INTO tbl VALUES (1, 'x'), (2, NULL)")

    print("current")
    run_all(cur)

    # cursor.py looks this up on the package at call time, so replacing the
    # attribute is enough to swap the transform.
    transforms.object_construct = patched

    print("\nwith the proposed fix")
    run_all(cur)

    print("\nlowered duckdb sql")
    for select in SELECTS:
        node = sqlglot.parse_one(f"SELECT {select} FROM tbl AS t", read="snowflake").selects[0]
        print(f"   {select:<33} -> {patched(node).sql(dialect='duckdb')}")
```

Local output:

```
./object_construct_star.py
current
   FAIL  object_construct(*)               ... Catalog Error: Scalar Function with name star_map does not exist!
   FAIL  object_construct_keep_null(*)     ... AssertionError (empty message)
   FAIL  object_construct(t.*)             ... Catalog Error: Scalar Function with name star_map does not exist!

with the proposed fix
   ok    object_construct(*)               -> ['{"A":1,"B":"x"}', '{"A":2}']
   ok    object_construct_keep_null(*)     -> ['{"A":1,"B":"x"}', '{"A":2,"B":null}']
   ok    object_construct(t.*)             -> ['{"A":1,"B":"x"}', '{"A":2}']

lowered duckdb sql
   object_construct(*)               -> JSON_MERGE_PATCH('{}', TO_JSON(STRUCT_PACK(*COLUMNS(*))))
   object_construct_keep_null(*)     -> TO_JSON(STRUCT_PACK(*COLUMNS(*)))
   object_construct(t.*)             -> JSON_MERGE_PATCH('{}', TO_JSON(t))
```

</details>

# Environment

```
fakesnow==0.11.15
duckdb==1.5.5
sqlglot==30.15.0
snowflake-connector-python==4.7.2
python 3.12.13
```

ishaan-upadhyay:
### Why?

Snowflake supports wildcard arguments for `OBJECT_CONSTRUCT` and related functions, but these queries currently fail in Fakesnow. This prevents Snowflake-compatible queries that construct objects from complete rows from running locally.

Fixes #394

### How?

Treat wildcard inputs as row values. For `OBJECT_CONSTRUCT`, a DuckDB macro serializes columns before unpivoting so SQL `NULL` values are omitted while explicit and nested JSON `null` values are preserved; `OBJECT_CONSTRUCT_KEEP_NULL` uses row serialization directly.

### DIFF
--- a/fakesnow/transforms/transforms.py
+++ b/fakesnow/transforms/transforms.py
@@ -874,16 +874,10 @@ def _star_object(star: Expr, *, keep_nulls: bool) -> Expr:
         # spread into STRUCT_PACK args so the column names don't need to be known here
         source = exp.Anonymous(this="STRUCT_PACK", expressions=[exp.var("*COLUMNS(*)")])
 
-    obj = exp.Anonymous(this="TO_JSON", expressions=[source])
     if keep_nulls:
-        return obj
+        return exp.Anonymous(this="TO_JSON", expressions=[source])
 
-    # OBJECT_CONSTRUCT omits nulls whereas TO_JSON keeps them. A JSON merge patch removes
-    # members whose value in the patch is null (RFC 7396), dropping exactly those.
-    return exp.Anonymous(
-        this="JSON_MERGE_PATCH",
-        expressions=[exp.Literal(this="{}", is_string=True), obj],
-    )
+    return exp.Anonymous(this="_FS_OBJECT_CONSTRUCT_STAR", expressions=[source])
 
 
 def object_construct(expression: Expr) -> Expr:
--- a/fakesnow/transforms/transforms.py
+++ b/fakesnow/transforms/transforms.py
@@ -845,12 +845,62 @@ def sample(expression: Expr) -> Expr:
     return expression
 
 
+# A star can carry a column filter, eg: * ILIKE 'col1%', which isn't supported yet.
+_STAR_FILTERS = ("ilike", "except_", "replace", "rename")
+
+
+def _star_arg(expression: Expr) -> Expr | None:
+    """Return the star argument of OBJECT_CONSTRUCT, if it's one that can be transformed."""
+
+    if not expression.is_star:
+        return None
+
+    star = expression.this if isinstance(expression, exp.Column) else expression
+    if any(star.args.get(name) for name in _STAR_FILTERS):
+        return None
+
+    return expression
+
+
+def _star_object(star: Expr, *, keep_nulls: bool) -> Expr:
+    """Build an object containing every column the star refers to."""
+
+    table = star.args.get("table") if isinstance(star, exp.Column) else None
+    if table is not None:
+        # a qualified star, eg: t.*, names a single source which TO_JSON accepts directly
+        source = exp.Column(this=table.copy())
+    else:
+        # an unqualified star is every column in scope, ie: whatever COLUMNS(*) expands to,
+        # spread into STRUCT_PACK args so the column names don't need to be known here
+        source = exp.Anonymous(this="STRUCT_PACK", expressions=[exp.var("*COLUMNS(*)")])
+
+    obj = exp.Anonymous(this="TO_JSON", expressions=[source])
+    if keep_nulls:
+        return obj
+
+    # OBJECT_CONSTRUCT omits nulls whereas TO_JSON keeps them. A JSON merge patch removes
+    # members whose value in the patch is null (RFC 7396), dropping exactly those.
+    return exp.Anonymous(
+        this="JSON_MERGE_PATCH",
+        expressions=[exp.Literal(this="{}", is_string=True), obj],
+    )
+
+
 def object_construct(expression: Expr) -> Expr:
-    """Convert OBJECT_CONSTRUCT/OBJECT_CONSTRUCT_KEEP_NULL to _FS_OBJECT_CONSTRUCT."""
+    """Convert OBJECT_CONSTRUCT/OBJECT_CONSTRUCT_KEEP_NULL to _FS_OBJECT_CONSTRUCT.
+
+    A star argument, eg: OBJECT_CONSTRUCT(*), becomes TO_JSON instead, because
+    _FS_OBJECT_CONSTRUCT needs the keys and values which a star doesn't provide.
+    """
 
     items: list[tuple[Expr, Expr]] = []
 
-    if isinstance(expression, exp.Struct):
+    if isinstance(expression, exp.StarMap):
+        # OBJECT_CONSTRUCT(*) or OBJECT_CONSTRUCT(t.*)
+        star = _star_arg(expression.this)
+        return _star_object(star, keep_nulls=False) if star else expression
+
+    elif isinstance(expression, exp.Struct):
         # OBJECT_CONSTRUCT
         keep_nulls = False
 
@@ -861,6 +911,10 @@ def object_construct(expression: Expr) -> Expr:
     elif isinstance(expression, exp.JSONObject):
         # OBJECT_CONSTRUCT_KEEP_NULL
         keep_nulls = True
+
+        # OBJECT_CONSTRUCT_KEEP_NULL(*) is a single star rather than key/value pairs
+        if len(expression.expressions) == 1 and (star := _star_arg(expression.expressions[0])):
+            return _star_object(star, keep_nulls=True)
 
         for kv in expression.expressions:
             assert isinstance(kv, exp.JSONKeyValue)
--- a/fakesnow/cursor.py
+++ b/fakesnow/cursor.py
@@ -473,7 +473,8 @@ class FakeSnowflakeCursor:
                 self._duck_conn.execute(sql, params)
         except duckdb.BinderException as e:
             msg = e.args[0]
-            raise snowflake.connector.errors.ProgrammingError(msg=msg, errno=2043, sqlstate="02000") from e
+            errno, sqlstate = (100103, "22000") if "Duplicate struct entry name" in msg else (2043, "02000")
+            raise snowflake.connector.errors.ProgrammingError(msg=msg, errno=errno, sqlstate=sqlstate) from e
         except duckdb.CatalogException as e:
             # minimal processing to make it look like a snowflake exception, message content may differ
             msg = cast(str, e.args[0]).split("\n")[0]
--- a/fakesnow/macros.py
+++ b/fakesnow/macros.py
@@ -49,6 +49,21 @@ CREATE OR REPLACE MACRO ${catalog}._fs_object_construct(keys, vals, keep_nulls) 
 """
 )
 
+FS_OBJECT_CONSTRUCT_STAR = Template(
+    """
+CREATE OR REPLACE MACRO ${catalog}._fs_object_construct_star(row_value) AS (
+    SELECT COALESCE(json_group_object(key, value::JSON), '{}'::JSON)
+    FROM (
+        UNPIVOT (
+            SELECT TO_JSON(COLUMNS(*))
+            FROM (SELECT UNNEST(row_value))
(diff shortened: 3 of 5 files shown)


fakesnow/transforms/transforms.py
```diff
@@ -845,12 +845,62 @@ def sample(expression: Expr) -> Expr:
     return expression
 
 
+# A star can carry a column filter, eg: * ILIKE 'col1%', which isn't supported yet.
```
ishaan-upadhyay, line 848: Ideally I'd like to provide this immediately, but there are a few tricky edge cases to consider, e.g.:

[Snowflake docs](https://docs.snowflake.com/en/sql-reference/functions/object_construct)

> For this function, the ILIKE and EXCLUDE keywords are valid only in a SELECT list or GROUP BY clause.

> The ILIKE and EXCLUDE keywords can’t be combined in a single function call.

In theory, this could be achieved using the DuckDB `SELECT` [syntax directly:](https://duckdb.org/docs/current/sql/query_syntax/select)

Image

Using `COLUMNS(regex)` branch of the grammar for `ILIKE` + `COLUMNS(... EXCLUDE)` for `EXCLUDE`. The ability to use the keywords in the `GROUP BY` clause in Snowflake as well may not be supported in DuckDB though.

At any rate, I can file a separate issue for this.

tekumara:
Thanks @ishaan-upadhyay for putting this together, especially the detailed issue, reproduction, and Snowflake comparison. I've pushed two follow-up commits to the branch:

- cad3d44 aligns the duplicate-key `ProgrammingError` with Snowflake's error code and SQLSTATE (`100103` / `22000`), so the same assertion works against Fakesnow and real Snowflake.
- 78bef8e preserves explicit and nested JSON `null` values in `OBJECT_CONSTRUCT(*)`. `JSON_MERGE_PATCH` also removes JSON nulls recursively, whereas Snowflake omits only SQL `NULL` values. The replacement macro serializes each column before unpivoting, allowing SQL nulls to be omitted without losing JSON nulls.

I verified all runtime tests added by this PR against real Snowflake, including the qualified-star and duplicate-key cases. The full local suite passes with 382 passed and 2 expected failures.

PR #398, opened 2026-08-26
feat: support more dtypes (more int widths, float, bool, category) in write_pandas auto_create_table

### LINKED ISSUE #397: Feature: Support additional dtypes (other int widths, floats, booleans) for write_pandas
# Summary

[`sql_type()`](https://github.com/tekumara/fakesnow/blob/main/fakesnow/pandas_tools.py) maps two dtypes and rejects the rest:

```python
def sql_type(dtype: np.dtype) -> str:
    if str(dtype) == "int64":
        return "NUMBER"
    elif str(dtype) == "object":
        return "VARCHAR"
    else:
        raise NotImplementedError(f"sql_type {dtype=}")
```

## What Snowflake produces under `write_pandas()`

Types read back with `DESCRIBE TABLE`: 

<details>
<summary>Table of dtype - Snowflake type - value - Fakesnow type</summary>

| dtype | Snowflake | value | fakesnow |
| --- | --- | --- | --- |
| `int64` | `NUMBER(38,0)` | `1` | `NUMBER(38,0)` |
| `int32` | `NUMBER(38,0)` | `1` | `NotImplementedError` |
| `Int64` (nullable) | `NUMBER(38,0)` | `1` | `NotImplementedError` |
| `float64` | `FLOAT` | `1.5` | `NotImplementedError` |
| `float32` | `FLOAT` | `1.5` | `NotImplementedError` |
| `bool` | `BOOLEAN` | `True` | `NotImplementedError` |
| `category` | `VARCHAR(16777216)` | `'a'` | `NotImplementedError` |
| `object` | `VARCHAR(16777216)` | `'a'` | `VARCHAR(16777216)` |

</details>

# Possible fix

Extending `sql_type()` covers it, mapping each dtype to what the table above shows as the Snowflake type for the respective Pandas `dtype`:

- all integer widths, signed/unsigned, plus the nullable `Int64` extension dtype → `NUMBER`
- `float32`/`float64` → `FLOAT`
- `bool` → `BOOLEAN`
- `category`, `string` → `VARCHAR`

# Reproducing

<details>

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "fakesnow>=0.11.15",
#   "duckdb==1.5.5",
#   "pandas",
# ]
# ///
"""write_pandas(auto_create_table=True) only accepts int64 and object columns.

Every dtype below round-trips on a real warehouse. Under fakesnow, anything
other than int64/object raises NotImplementedError from sql_type().
"""

from __future__ import annotations

import duckdb
import fakesnow
import pandas as pd
import snowflake.connector
from fakesnow.pandas_tools import write_pandas

COLUMNS = {
    "int64": pd.Series([1], dtype="int64"),
    "int32": pd.Series([1], dtype="int32"),
    "float64": pd.Series([1.5], dtype="float64"),
    "bool": pd.Series([True], dtype="bool"),
    "category": pd.Series(["a"], dtype="category"),
    "object": pd.Series(["a"], dtype="object"),
    "Int64 (nullable)": pd.Series([1], dtype="Int64"),
}


def main() -> None:
            account="fakesnow", user="u", password="p", database="TESTDB", schema="PUBLIC"
        )
        cur = conn.cursor()

        for i, (label, series) in enumerate(COLUMNS.items()):
            df = pd.DataFrame({"c": series})
            table = f"T{i}"
            try:
                write_pandas(conn, df, table, auto_create_table=True, quote_identifiers=False)
            except Exception as e:
                print(f"  {label:<20} {type(e).__name__}: {e}")
                continue
            cur.execute(f"describe table {table}")
            live = cur.fetchone()[1]
            cur.execute(f"select c from {table}")
            print(f"  {label:<20} {live:<20} {cur.fetchone()[0]!r}")

    print(f"\n  fakesnow==0.11.15 duckdb=={duckdb.__version__} pandas=={pd.__version__}")


if __name__ == "__main__":
    main()
```

```
  int64                NUMBER(38,0)         1
  int32                NotImplementedError: sql_type dtype=dtype('int32')
  float64              NotImplementedError: sql_type dtype=dtype('float64')
  bool                 NotImplementedError: sql_type dtype=dtype('bool')
  category             NotImplementedError: sql_type dtype=CategoricalDtype(categories=['a'], ordered=False, categories_dtype=str)
  object               VARCHAR(16777216)    'a'
  Int64 (nullable)     NotImplementedError: sql_type dtype=Int64Dtype()

  fakesnow==0.11.15 duckdb==1.5.5 pandas==3.0.5
```

</details>

The Snowflake columns come from the same frames written to a real warehouse with `snowflake.connector.pandas_tools.write_pandas`.

# Environment

```
fakesnow==0.11.15
duckdb==1.5.5
pandas==3.0.5
snowflake-connector-python==4.7.2
python 3.13.12
```

## Future work

- Supporting `use_logical_type=True` in a limited capacity, which I will file a follow-up issue for. For now it seems the simpler path is to only support timestamps when this is set, so we can rely on DuckDB's direct scanning of DataFrame types.

ishaan-upadhyay:
Fixes #397 (full details there)

`sql_type` currently maps only `int64` and `object`, so `write_pandas(auto_create_table=True)` could not create a table from a frame holding a float, bool, narrower/nullable int, or category column. 

Special cases:

- A category maps to the type of its categories, since that's what the staged parquet dictionary holds.
- `float16` is stored as binary

Verified against a live warehouse: `snowflake.connector.pandas_tools.write_pandas` creates the same column types as this change for the frame in `test_write_pandas_auto_create_dtypes`.

Made with [Cursor](https://cursor.com)


fakesnow/pandas_tools.py
```diff
@@ -110,6 +136,9 @@ def _insert_df(duck_conn: DuckDBPyConnection, df: pd.DataFrame, table_name: str)
         df[col] = df[col].apply(lambda x: json.dumps(x) if isinstance(x, (dict, list)) else x)
 
     escaped_cols = ",".join(f'"{col}"' for col in df.columns.to_list())
-    duck_conn.execute(f"INSERT INTO {table_name}({escaped_cols}) SELECT * FROM df")
+    # duckdb reads a temporal column as a TIMESTAMP or INTERVAL and casts neither to a number,
+    # so nanoseconds since the epoch are selected explicitly.
+    values = ",".join(f'epoch_ns("{col}")' if col in epoch_ns_cols else f'"{col}"' for col in df.columns.to_list())
+    duck_conn.execute(f"INSERT INTO {table_name}({escaped_cols}) SELECT {values} FROM df")
```
ishaan-upadhyay: I will investigate this more before marking ready for review. I have doubts that it's preferable to stuff more and more workarounds into `_insert_df()`. 

It may be preferable to instead cast the column on the DataFrame itself (may require making a copy to avoid mutating the source DF).

ishaan-upadhyay: Dropped from this PR

fakesnow/pandas_tools.py
```diff
@@ -35,11 +35,21 @@
...
+def sql_type(dtype: np.dtype | pd.api.extensions.ExtensionDtype) -> str:
+    import pandas as pd
+
+    if pd.api.types.is_bool_dtype(dtype):
+        return "BOOLEAN"
+    elif pd.api.types.is_integer_dtype(dtype):
         return "NUMBER"
-    elif str(dtype) == "object":
+    elif pd.api.types.is_float_dtype(dtype) and str(dtype) != "float16":
+        # snowflake reads a staged half float as binary, not as a float
```
ishaan-upadhyay: `float16` is a [Parquet logical type](https://parquet.apache.org/docs/file-format/types/logicaltypes/#:~:text=FLOAT16):

> The primitive type is a 2-byte FIXED_LEN_BYTE_ARRAY.

Which is why Snowflake is preserving it as `binary` 

Not sure if it's preferable to fail here, or serialize as DuckDB `BINARY`. DuckDB would treat the DataFrame column as `FLOAT` so this would require modifying `_insert_df` to cast any such column.

(I would argue this is an upstream bug with Snowflake. `USE_LOGICAL_TYPE=TRUE` on `snowflake-connector-python` `write_pandas` outright fails here too).

ishaan-upadhyay:
(Also, I apologize for the commit amends, I know it makes PRs harder to review and track incremental changes. Will ensure this is not the case in the future)

fakesnow/pandas_tools.py
tekumara: I've pushed fddd3eb to support float16 as it doesn't add much complexity to do so.

tekumara:
Thanks @ishaan-upadhyay FYI I've pushed 6b352ac to improve the test assertions.

tekumara (approved):

Issue #258, opened 2025-08-18
Add support for mocking `COPY INTO` with Parquet output in `FakeSnowflakeCursor`


zetaattopsortcom:
In my project I needed to mock `COPY INTO` statements so that queries could produce Parquet files in a local stage path. To achieve this, I extended `fakesnow.cursor.FakeSnowflakeCursor` and overrode `execute` as follows:

```python
    def execute(self, command: str, params: Sequence[Any] | dict[Any, Any] | None = None) -> Self:
        if self._stage_path is not None:
            formatted_command = command % params if params else command
            parsed_command = sqlglot.parse_one(formatted_command, read='snowflake')
            if isinstance(parsed_command, sqlglot.expressions.Copy):
                inner_sql = parsed_command.args['files'][0].this
                self._cursor.execute(inner_sql.sql(dialect='snowflake'), params)
                self._write_parquet(parsed_command, self._stage_path)
                self._results = []
                return self
        self._cursor.execute(command, params)
        self._results = self._cursor.fetchall()
        return self
```

The _write_parquet helper inspects the COPY INTO parameters and writes the results into a Parquet file. Only Parquet is supported today, but it could be expanded in a similar way to handle other formats:

```python
    def _write_parquet(
        self, copy_command: sqlglot.expressions.Copy, stage_path: pathlib.Path
    ) -> None:
        stage_target = str(copy_command.this.this.this).lstrip('@')

        file_type = 'parquet'
        compression: _PARQUET_COMPRESSION_TYPES = 'snappy'

        for param in copy_command.args.get('params', []):
            if str(param.this.this).upper() == 'FILE_FORMAT':
                for expr in param.args.get('expressions', []):
                    prop_name = str(expr.this.this).upper()
                    prop_value = str(expr.args['value'].this).strip('"\'')
                    if prop_name == 'TYPE':
                        file_type = prop_value.lower()
                    elif prop_name == 'COMPRESSION':
                        compression = typing.cast(_PARQUET_COMPRESSION_TYPES, prop_value.lower())

        if file_type != 'parquet':
            raise ValueError(f'Only parquet is supported by the mock. Got TYPE={file_type!r}')

        columns = (
            [d[0] for d in self._cursor.description] if self._cursor.description else ['_COL_0']
        )
        fetched_rows = self._cursor.fetchall()
        if fetched_rows:
            if self._cursor._use_dict_result:
                rows = [
                    tuple(r.get(c) for c in columns)
[...]

PR #400, opened 2026-08-28
feat: support a stage/PUT/COPY/MERGE bulk-load pipeline in server mode

### LINKED ISSUE #403: Server mode cannot run a stage/PUT/COPY/MERGE bulk-load pipeline
## Context

We run fakesnow in server mode (docker) as a local Snowflake emulator for the acceptance/integration tests of a Postgres → Snowflake data-sync service. The service uses the classic bulk-load pattern: create a stage and a named file format, `PUT` a gzipped CSV, `COPY INTO` a staging table, then `MERGE` into the target. Table snapshots additionally use table stages (`@db.schema.%table`).

Verified against fakesnow 0.11.15, every step of that pipeline fails:

## Problems

1. **`CREATE OR REPLACE STAGE` / `CREATE STAGE IF NOT EXISTS` are not honored** — re-creating an existing stage always fails with `Object '<name>' already exists` (errno 2002), so any job that recreates its stage per run cannot execute twice.

2. **`CREATE FILE FORMAT` crashes the server** — `fakesnow/checks.py` `_missing_qualifiers` raises `AssertionError: Unexpected parent kind: FILE FORMAT`, returning HTTP 500. The snowflake connector then retries the 500 in a loop, so the client hangs for ~10 minutes instead of failing fast.

3. **`COPY INTO ... FILE_FORMAT = (FORMAT_NAME = '<name>')` is not supported** — fails with `FILE_FORMAT without TYPE is not currently implemented` (follows from 2: named formats don't exist).

4. **Common CSV copy options are not implemented** — `EMPTY_FIELD_AS_NULL`, `NULL_IF`, `ESCAPE_UNENCLOSED_FIELD='NONE'` fail with "not currently implemented", although they map naturally onto duckdb `read_csv` defaults.

5. **`PUT` with an unquoted file URL fails to parse** — snowflake-sqlalchemy sends `PUT file:///path/data.csv.gz @STAGE AUTO_COMPRESS=FALSE` (unquoted); sqlglot only recognizes `exp.Put` with a quoted source, so the statement falls through untransformed to duckdb and dies with `Parser Error: syntax error at or near "PUT"`.

6. **Server-mode `PUT` writes into the client's filesystem** — the server returns `stageInfo.locationType = LOCAL_FS` with a server-local path, so the connector writes the file on the *client* machine. When client and server are different containers (the normal docker setup), a subsequent `COPY INTO` finds no files and silently loads 0 rows.

7. **Table stages (`@db.schema.%table`) are not supported** for `PUT`/`COPY INTO`.

8. **Multi-clause `MERGE` returns `NULL` counts** — with `WHEN MATCHED ... UPDATE` + `WHEN NOT MATCHED ... INSERT`, a clause that affects 0 rows yields `None` instead of `0` in the result row ("number of rows inserted"/"number of rows updated"); snowflake-connector crashes client-side with `TypeError: int() argument must be ... not 'NoneType'`. Real Snowflake always returns integers.

## Repro (condensed)

```sql
CREATE OR REPLACE STAGE MY_STAGE FILE_FORMAT = (TYPE = 'CSV');           -- fails on 2nd run (1)
CREATE OR REPLACE FILE FORMAT MY_FMT TYPE='CSV' SKIP_HEADER=1;           -- HTTP 500 (2)
PUT file:///tmp/data.csv.gz @MY_STAGE AUTO_COMPRESS=FALSE;               -- parser error (5), wrong filesystem (6)
COPY INTO STAGING FROM @MY_STAGE/data.csv.gz
  FILE_FORMAT = (FORMAT_NAME = 'MY_FMT');                                -- not implemented (3)
COPY INTO STAGING FROM @MY_STAGE/data.csv.gz
  FILE_FORMAT = (TYPE='CSV' EMPTY_FIELD_AS_NULL=TRUE NULL_IF=(''));      -- not implemented (4)
PUT file:///tmp/snap.csv.gz @DB.SCH.%SNAP_T AUTO_COMPRESS=FALSE;         -- not supported (7)
MERGE INTO T USING S ON ... WHEN MATCHED THEN UPDATE ...
  WHEN NOT MATCHED THEN INSERT ...;                                      -- None counts (8)
```

PR #400 implements all of the above.

arnaldojvg:
Implements the Snowflake features needed to run acceptance tests of a stage/PUT/COPY/MERGE bulk-load pipeline (Postgres → Snowflake data-sync job using snowflake-sqlalchemy) against fakesnow in server mode.

Fixes #403

## Features

- **CREATE OR REPLACE STAGE** replaces an existing stage (and empties its internal storage), and **CREATE STAGE IF NOT EXISTS** succeeds without error when the stage exists.
- **CREATE [OR REPLACE] FILE FORMAT [IF NOT EXISTS]**: named file formats are stored in the info schema (`_fs_global._fs_information_schema._fs_file_formats`). Previously this crashed the server with a 500 (`AssertionError: Unexpected parent kind: FILE FORMAT` in checks.py).
- **COPY INTO ... FILE_FORMAT = (FORMAT_NAME = '...')** resolves a named file format to its stored options. Previously: "FILE_FORMAT without TYPE is not currently implemented".
- **More CSV options in COPY INTO / file formats**, mapped onto duckdb's `read_csv`: `SKIP_HEADER=<n>`, `NULL_IF`, `EMPTY_FIELD_AS_NULL`, `ESCAPE_UNENCLOSED_FIELD='NONE'`, `COMPRESSION` (AUTO/GZIP/NONE). `ON_ERROR` also accepts a string literal (`ON_ERROR = 'ABORT_STATEMENT'`).
- **PUT with an unquoted file url** (as sent by eg: snowflake-sqlalchemy) now works: sqlglot only parses PUT as `exp.Put` when the source is a quoted literal, so the statement fell through to duckdb raw and failed with `Parser Error: syntax error at or near "PUT"`. The source is now quoted before parsing (both embedded and server mode).
- **PUT options**: `AUTO_COMPRESS`, `OVERWRITE`, `PARALLEL`, `SOURCE_COMPRESSION` are honoured (and returned to the client in server mode, so the connector's file transfer agent respects them). An already gzipped source is no longer recompressed.
- **COPY from a stage path suffix**, eg: `COPY INTO t FROM @stage/file.csv.gz` (prefix match, like Snowflake).
- **Table stages** for PUT and COPY, eg: `PUT file://data.csv @db.schema.%mytable` and `COPY INTO t FROM @db.schema.%mytable/data.csv.gz`.
- **Load history is scoped to the target table**: the same file loaded into a different table is not skipped, and recreating a table (eg: `CREATE OR REPLACE TEMPORARY TABLE staging`) resets its load history, matching Snowflake's per-table load metadata.

An end-to-end server-mode test runs the full pipeline (CREATE OR REPLACE STAGE / FILE FORMAT / TEMPORARY TABLE, PUT AUTO_COMPRESS=FALSE, COPY with FORMAT_NAME, MERGE) twice on one server to prove reruns work.

## Out of scope

- `LIST @%table_stage` — sqlglot cannot parse LIST with a table stage reference; table stages are supported for PUT and COPY only.
- `ESCAPE_UNENCLOSED_FIELD` values other than `'NONE'`, and `EMPTY_FIELD_AS_NULL=FALSE` combined with `NULL_IF=('')`, raise NotImplementedError as they have no faithful duckdb mapping.
- `SHOW FILE FORMATS` / `DROP FILE FORMAT`.

### DIFF
--- a/fakesnow/checks.py
+++ b/fakesnow/checks.py
@@ -84,8 +84,8 @@ def _missing_qualifiers(node: exp.Table) -> tuple[bool, bool]:
             # "CREATE/DROP SCHEMA"
             no_database = not node.args.get("catalog")
             no_schema = False
-        elif parent_kind.upper() in {"TABLE", "VIEW", "STAGE", "TAG", "SEQUENCE"}:
-            # "CREATE/DROP TABLE/VIEW/STAGE/TAG"
+        elif parent_kind.upper() in {"TABLE", "VIEW", "STAGE", "TAG", "SEQUENCE", "FILE FORMAT"}:
+            # "CREATE/DROP TABLE/VIEW/STAGE/TAG/FILE FORMAT"
             no_database = not node.args.get("catalog")
             no_schema = not node.args.get("db")
         else:
--- a/fakesnow/copy_into.py
+++ b/fakesnow/copy_into.py
@@ -17,6 +17,7 @@ from sqlglot import Expr, exp
 import fakesnow.transforms.stage as stage
 from fakesnow import logger
 from fakesnow.params import MutableParams, pop_qmark_param
+from fakesnow.transforms.file_format import format_options, lookup_file_format
 
 Params = Sequence[Any] | dict[Any, Any]
 
@@ -46,7 +47,9 @@ def copy_into(
     expr: exp.Copy,
     params: MutableParams | None = None,
 ) -> str:
-    cparams = _params(expr, params)
+    cparams = _params(
+        expr, params, duck_conn=duck_conn, current_database=current_database, current_schema=current_schema
+    )
 
     from_source = _from_source(expr)
     source = (
@@ -203,7 +206,13 @@ def _get_table_columns(
     return [col[0].upper() for col in columns]
 
 
-def _params(expr: exp.Copy, params: MutableParams | None = None) -> CopyParams:
+def _params(
+    expr: exp.Copy,
+    params: MutableParams | None = None,
+    duck_conn: DuckDBPyConnection | None = None,
+    current_database: str | None = None,
+    current_schema: str | None = None,
+) -> CopyParams:
     kwargs = {}
     force = False
     purge = False
@@ -217,12 +226,16 @@ def _params(expr: exp.Copy, params: MutableParams | None = None) -> CopyParams:
             if kwargs.get("file_format"):
                 raise ValueError(cparams)
 
-            var_type = next((e.args["value"].this for e in param.expressions if e.this.this == "TYPE"), None)
-            if not var_type:
-                raise NotImplementedError("FILE_FORMAT without TYPE is not currently implemented")
+            options = format_options(list(param.expressions))
+            if format_name := options.pop("FORMAT_NAME", None):
+                if options.get("TYPE"):
+                    raise ValueError("Cannot specify both FORMAT_NAME and TYPE in FILE_FORMAT")
+                assert duck_conn, "duck_conn is required to resolve FORMAT_NAME"
+                options = lookup_file_format(duck_conn, str(format_name), current_database, current_schema)
 
+            var_type = str(options.get("TYPE", "CSV")).upper()
             if var_type == "CSV":
-                kwargs["file_format"] = handle_csv(param.expressions)
+                kwargs["file_format"] = handle_csv(options)
             elif var_type == "PARQUET":
                 kwargs["file_format"] = ReadParquet()
             else:
@@ -234,7 +247,7 @@ def _params(expr: exp.Copy, params: MutableParams | None = None) -> CopyParams:
         elif var == "PURGE":
             purge = True
         elif var == "ON_ERROR":
-            if isinstance(param.expression, exp.Var):
+            if isinstance(param.expression, (exp.Var, exp.Literal)):
                 on_error = param.expression.name.upper()
             elif isinstance(param.expression, exp.Placeholder):
                 on_error = pop_qmark_param(params, expr, param.expression)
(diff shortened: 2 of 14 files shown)


arnaldojvg:
related:
https://github.com/tekumara/fakesnow/issues/258

tekumara:
thanks for the contribution @arnaldojvg - fyi this is a large PR and will take me some time to review each commit, if you wanted to progress some of these more quickly then I recommend raising individual PRs

arnaldojvg:
Thanks @tekumara, let me know of anything 🙏

arnaldojvg:
Thanks for the review and the suggestion — agreed this is a lot to review at once. I've split it into four issues and PRs, all cherry-picked verbatim from this branch, each independently green (full suite + hooks):

| Issue | PR | Scope | Depends on |
|-------|----|-------|------------|
| #406 | #410 | fix: MERGE returns integer counts when no rows are affected | — |
| #407 | #411 | fix: scope load history to the target table | — |
| #408 | #412 | feat: CREATE OR REPLACE STAGE, named file formats, and more CSV options in COPY INTO | — |
| #409 | #413 | feat: PUT over http in server mode, unquoted PUT urls, PUT options, and table stages (+ end-to-end pipeline test) | #412, #411 |

#410, #411 and #412 are independent and can be reviewed in any order; #413 is based on #412's branch and carries #411's commit until they merge, so it is green standalone and its diff shrinks to its own changes as the others land.

This PR remains as the integrated reference — happy for it to be closed in favour of the splits whenever you prefer.

arnaldojvg:
I'll close this since we are managing this on the smaller PRs

PR #402, opened 2026-08-31
fix: report the rows affected by a MERGE as the total across its operations

anjeongkyun:
In local mode `cursor.rowcount` is 1 for every merge, whatever it did. The merge result is a single row holding a count per operation, and that row was counted rather than the counts in it.

An account reports the total across the operations:

| merge | result row | rowcount |
| --- | --- | --- |
| matched update + not matched insert | `(1, 1)` | 2 |
| matched delete | `(1,)` | 1 |
| update + insert + delete | `(1, 1, 0)` | 2 |

Marked the statement that produces those counts in the merge transform, the same way the other transforms flag statements they need the cursor to treat specially, and summed them there.

The server was already correct because the connector sums the counts from the rowset itself, which `test_server_rowcount_dml` covers. This is local mode only.

One thing worth noting: the sum comes back as a `Decimal`, and returning that as `rowcount` broke the server's JSON response, so it's cast to `int`. `test_server_rowcount_dml` caught it.

### DIFF
--- a/fakesnow/cursor.py
+++ b/fakesnow/cursor.py
@@ -622,6 +622,10 @@ class FakeSnowflakeCursor:
             self._duck_conn.execute(result_sql)
 
         self._arrow_table = self._duck_conn.to_arrow_table()
+        if transformed.args.get("merge_counts"):
+            # a merge returns a count per operation, but snowflake reports their total as the
+            # rows affected, not the single row those counts are in
+            affected_count = int(sum(c[0].as_py() for c in self._arrow_table.columns))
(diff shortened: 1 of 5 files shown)


tekumara (commented):
Thanks @anjeongkyun for the focused MERGE rowcount fix.

I pushed two follow-up commits:

- 3bbc44e — use the public `cur.rowcount` API in the server.
- ba172f4 — remove `test_server_rowcount_dml`: the shared-cursor `test_merge_rowcount` now covers the rowcount behavior, and the server test still passed when the historic server-specific implementation was removed.

tekumara (approved):

PR #410, opened 2026-09-02
fix: MERGE returns integer counts when no rows are affected

### LINKED ISSUE #406: Multi-clause MERGE returns NULL counts when no rows are affected
With a multi-clause MERGE, eg:

```sql
MERGE INTO t USING s ON t.id = s.id
  WHEN MATCHED AND t.name IS DISTINCT FROM s.name THEN UPDATE SET name = s.name
  WHEN NOT MATCHED THEN INSERT (id, name) VALUES (s.id, s.name);
```

when no WHEN clause affects any row (eg: re-running the merge when source equals target), the result row is `(None, None)` instead of `(0, 0)`. Real Snowflake always returns integer counts.

In server mode, snowflake-connector-python reads the counts client-side and crashes:

```
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

Cause: the counts are computed with duckdb's `COUNT_IF`, which is SUM-based and returns NULL over an empty `merge_candidates` table.

arnaldojvg:
duckdb's `COUNT_IF` returns NULL over an empty `merge_candidates` table, which crashed clients that read the counts, eg: snowflake-connector-python calls `int()` on them. Snowflake always returns integers, so wrap each count in `COALESCE(..., 0)`.

Regression tests cover both local mode and the server/connector path (where the `int()` crash happened).

Fixes #406. Split out of #400 per maintainer request.

### DIFF
--- a/fakesnow/cursor.py
+++ b/fakesnow/cursor.py
@@ -626,7 +626,7 @@ class FakeSnowflakeCursor:
             # a merge returns a count per operation, but snowflake reports their total as the
             # rows affected, not the single row those counts are in
             affected_count = int(sum(c[0].as_py() for c in self._arrow_table.columns))
-        self._rowcount = affected_count or self._arrow_table.num_rows
+        self._rowcount = affected_count if affected_count is not None else self._arrow_table.num_rows
         self._sfqid = str(uuid.uuid4())
 
         self._last_sql = result_sql or sql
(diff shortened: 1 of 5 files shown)


tekumara:
Thanks for the contribution @arnaldojvg - fyi I've pushed some tweaks to this branch to tighten up result row counts.

tekumara (approved):

PR #411, opened 2026-09-02
fix: scope load history to the target table

### LINKED ISSUE #407: COPY INTO load-history dedup is global instead of per table
fakesnow skips a COPY of a file that appears in `_fs_load_history`, but the check is by file name only, across all tables. Snowflake tracks load metadata per table:

- the same file copied into a *different* table is loaded, not skipped
- recreating a table (eg: `CREATE OR REPLACE TEMPORARY TABLE staging ...`) resets its load history, so a bulk-load job that recreates its staging table each run loads the file again

Observed: the second run of a stage/PUT/COPY pipeline silently returns `LOAD_SKIPPED` / "File was loaded before." and loads 0 rows even though the staging table was just recreated.

```sql
CREATE OR REPLACE TEMPORARY TABLE staging (...);
COPY INTO staging FROM @my_stage/data.csv.gz ...;   -- LOAD_SKIPPED on the 2nd pipeline run
```

arnaldojvg:
Snowflake tracks load metadata per table, so the same file loaded into a different table is not skipped, and recreating a table resets its load history.

The dedup check now matches on file name, table name and schema, and `CREATE TABLE` clears the table's load-history rows.

Fixes #407. Split out of #400 per maintainer request.


tekumara (approved):
thank you!

PR #412, opened 2026-09-02
feat: CREATE OR REPLACE STAGE, named file formats, and more CSV options in COPY INTO

arnaldojvg:
Note: #413 is stacked on this branch — please merge this one first.

- **Stages**: `_fs_stages` gains a primary key so OR REPLACE maps to duckdb `INSERT OR REPLACE`; OR REPLACE also empties the stage's backing storage, as in Snowflake; IF NOT EXISTS returns "already exists, statement succeeded." instead of erroring.
- **Named file formats**: stored (with their options as JSON) in a new `_fs_global._fs_information_schema._fs_file_formats` table; the FILE FORMAT kind is accepted in checks.py instead of asserting (which returned HTTP 500 in server mode); OR REPLACE and IF NOT EXISTS supported.
- **COPY INTO**: `FILE_FORMAT = (FORMAT_NAME = ...)` resolves stored formats; `SKIP_HEADER > 1`, `NULL_IF`, `EMPTY_FIELD_AS_NULL`, `ESCAPE_UNENCLOSED_FIELD='NONE'` and `COMPRESSION` map onto duckdb's `read_csv` (combinations with no faithful duckdb mapping raise NotImplementedError); `ON_ERROR` accepts a string literal; a stage reference can include a path suffix, eg: `@stage1/file.csv.gz` (prefix match, as in Snowflake).

Fixes #408. Split out of #400 per maintainer request.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #413, opened 2026-09-02
feat: PUT over http in server mode, unquoted PUT urls, PUT options, and table stages

### LINKED ISSUE #409: Server-mode PUT is unusable: unquoted urls fail to parse, files land on the client, table stages unsupported
Three related gaps that block the PUT half of a bulk-load pipeline (verified on 0.11.15):

1. **PUT with an unquoted file url fails with a duckdb parser error** — snowflake-sqlalchemy (and plain client code) sends the source unquoted; sqlglot only recognises `exp.Put` with a quoted literal, so the statement falls through untransformed to duckdb:

```sql
PUT file:///tmp/data.csv.gz @MY_STAGE AUTO_COMPRESS=FALSE;
-- 001003 (42000): Parser Error: syntax error at or near "PUT"
```

PUT options are also ignored: `AUTO_COMPRESS=FALSE` still gzips (and renames) the file, and an already gzipped source is recompressed.

2. **Server-mode PUT writes into the client's filesystem** — the PUT response returns `stageInfo.locationType = LOCAL_FS` with a server-local path, so the connector writes the staged file on the *client* machine. When client and server don't share a filesystem — the normal docker setup — a subsequent COPY INTO finds no files and reports "Copy executed with 0 files processed": 0 rows loaded, no error anywhere.

3. **Table stages (`@db.schema.%table`) are not supported** for PUT/COPY INTO:

```sql
PUT file:///tmp/snap.csv.gz @DB.SCH.%SNAP_T AUTO_COMPRESS=FALSE;
-- Stage 'DB.SCH.%SNAP_T' does not exist or not authorized.
```

arnaldojvg:
Depends on #412 (based on its branch) — contains its commits until it merges. Please merge #412 first; I'll rebase this PR right after so only its own 4 commits remain. (If this one is merged first instead, #412 can simply be closed as already included.)

- **Unquoted PUT urls**: sqlglot only parses PUT as `exp.Put` when the source is a quoted literal, so an unquoted url (as sent by eg: snowflake-sqlalchemy) fell through to duckdb as a raw command; the source is quoted before parsing. `AUTO_COMPRESS`, `OVERWRITE`, `PARALLEL` and `SOURCE_COMPRESSION` are honoured, and an already gzipped source is no longer recompressed.
- **PUT over http in server mode**: the server returned a LOCAL_FS stage location, so the connector wrote the file to the client's own filesystem; when the client doesn't share the server's filesystem (eg: the server runs in a container) a subsequent COPY INTO silently found no files. The server now returns a GCS-style stage info with a presigned url pointing back at the fakesnow server, so the connector uploads the file over http into the stage's backing storage (GCS is the only location type the connector uploads to via a plain http presigned url). A PUT with a `?` placeholder target keeps LOCAL_FS because the connector re-requests the presigned url by executing the command without bindings.
- **Table stages**: a stage name starting with `%` is a table stage, which exists implicitly for every table: existence is checked against the table catalog rather than the stage registry, and its files live in the internal-stage backing storage. Out of scope: `LIST @%table`, which sqlglot cannot parse.
- Includes an end-to-end server-mode test running the full stage/PUT/COPY/MERGE pipeline twice on one server (the rerun needs #411's per-table load history, hence the carried commit).

Fixes #409. Split out of #400 per maintainer request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,6 +305,7 @@ Partially supported:
 - Tags
 - User management
 - Stages and PUT
+- Named file formats
 - `COPY INTO` from S3 sources and stages, see [COPY INTO](#copy-into)
 
 Not yet implemented:
--- a/fakesnow/checks.py
+++ b/fakesnow/checks.py
@@ -84,8 +84,8 @@ def _missing_qualifiers(node: exp.Table) -> tuple[bool, bool]:
             # "CREATE/DROP SCHEMA"
             no_database = not node.args.get("catalog")
             no_schema = False
-        elif parent_kind.upper() in {"TABLE", "VIEW", "STAGE", "TAG", "SEQUENCE"}:
-            # "CREATE/DROP TABLE/VIEW/STAGE/TAG"
+        elif parent_kind.upper() in {"TABLE", "VIEW", "STAGE", "TAG", "SEQUENCE", "FILE FORMAT"}:
+            # "CREATE/DROP TABLE/VIEW/STAGE/TAG/FILE FORMAT"
             no_database = not node.args.get("catalog")
             no_schema = not node.args.get("db")
         else:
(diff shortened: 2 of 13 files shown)


arnaldojvg:
Superseded by #420, #422 and #424 (split per the discussion in #412); closing.

PR #420, opened 2026-09-10
fix: PUT with an unquoted file url and support PUT options

arnaldojvg:
Accepts `PUT` commands the way the connector and most Snowflake clients issue them. An unquoted `file://` source is quoted before parsing, because sqlglot otherwise treats the statement as an opaque `Command` and fakesnow never recognises it as a PUT (this applies both in-process and in server mode). PUT options are honoured: `AUTO_COMPRESS = FALSE` and already-gzipped sources are stored as-is instead of being gzipped again, and `OVERWRITE`, `SOURCE_COMPRESSION` and `PARALLEL` are passed through to the connector. The upload result now reports the real source and target compression.

Tests: `test_put_unquoted_src_auto_compress_false`, `test_put_gzipped_src_not_recompressed` in `tests/test_stage.py`.


tekumara:
Hi @arnaldojvg this PR and the other  ones have conflicts with main

fakesnow/transforms/stage.py
```diff
@@ -205,11 +236,11 @@ def put_stage(
...
-        # defaults as per https://docs.snowflake.com/en/sql-reference/sql/put TODO: support other values
-        "parallel": 4,
-        "autoCompress": True,
-        "sourceCompression": "auto_detect",
-        "overwrite": False,
+        # defaults as per https://docs.snowflake.com/en/sql-reference/sql/put
+        "parallel": options.get("PARALLEL", 4),
+        "autoCompress": options.get("AUTO_COMPRESS", True),
+        "sourceCompression": str(options.get("SOURCE_COMPRESSION", "auto_detect")).lower(),
+        "overwrite": options.get("OVERWRITE", False),
```
tekumara, line 243: only `AUTO_COMPRESS` has been implemented in direct (ie: non-server) fakesnow .. 

what do you need these PUT options for?

arnaldojvg: We only need `AUTO_COMPRESS=FALSE`: our job PUTs an already gzipped csv, so with the hardcoded `autoCompress: True` the connector gzips it again and the staged name/contents no longer match what `COPY INTO` expects. The other three only replaced the hardcoded defaults (the `TODO: support other values` comment) with pass-through of what the client sent; in server mode the connector's file transfer agent reads `parallel`/`sourceCompression`/`overwrite` from this response, in direct mode only `autoCompress` is consumed.

I've opened #428 as a trimmed alternative that honours only `AUTO_COMPRESS` and keeps the other defaults. Pick whichever you prefer and I'll close the other and rebase #422 / #424 onto it.

tekumara: Thanks for raising #428  - lets use that instead as the scope is smaller and fully tested.

PR #422, opened 2026-09-10
fix: transfer PUT files over http in server mode

arnaldojvg:
In server mode the PUT response told the connector to write the file to a `LOCAL_FS` path, which only works when the client shares the server's filesystem and breaks when fakesnow runs in a container. The server now returns a GCS-style `stageInfo` with a presigned url pointing at a new `PUT /fs_bucket/{path}` route, so the connector uploads the file over http and the server stores it in the stage's backing directory. GCS is the only location type the connector uploads to via a plain http url, and it re-requests the url with the destination file name (`file://data.csv.gz`), so `put_stage` now includes the url netloc in the source path. A PUT with a bound `?` target keeps `LOCAL_FS`, because the connector re-runs that command without bindings.

Tests: `test_server_put_presigned_url`, `test_server_put_qmark_target_stays_local`, `test_server_bucket_upload_rejects_path_outside_bucket` in `tests/test_server.py`. The existing server-mode test `tests/test_copy_into.py::test_copy_internal_stage_server` also goes through the http upload.


tekumara:
@arnaldojvg FYI this now has merge conflicts

arnaldojvg:
Rebased onto main, all three (#422, #423, #424) are clean again. #423 also adapts to #431: the FORMAT_NAME lookup now strips the stored defaults so they don't reach COPY.

PR #424, opened 2026-09-10
feat: support table stages and staged file paths

arnaldojvg:
Depends on #422 and #423

A table stage (`@%table` or `@db.schema.%table`) exists implicitly for every table, so `PUT`, `LIST` and `COPY INTO` now resolve it against the table catalog instead of the stages table, and a PUT to a table stage for a missing table raises Snowflake's `Stage ... does not exist or not authorized`. A stage reference can also carry a path suffix (`@stage/dir/file.csv.gz`), which `COPY INTO` treats as a prefix match within the stage, as Snowflake does. Internal stage globs use the plain path rather than a `file://` uri because duckdb does not decode percent-encoded uris.

This is the last slice, so it also adds the end-to-end server-mode tests that need everything above: the stage/PUT/COPY/MERGE bulk-load pipeline run twice (`test_server_bulk_load_pipeline`) and loading a gzipped csv with VARIANT columns from a fully qualified table stage (`test_server_table_stage_bulk_load`).

Tests: `test_copy_internal_stage_path`, `test_copy_internal_table_stage`, `test_put_table_stage_non_existent_table` in `tests/test_copy_into.py`; `test_server_bulk_load_pipeline`, `test_server_table_stage_bulk_load` in `tests/test_server.py`.


tekumara:
@arnaldojvg FYI this now has merge conflicts

arnaldojvg:
Rebased onto main, all three (#422, #423, #424) are clean again. #423 also adapts to #431: the FORMAT_NAME lookup now strips the stored defaults so they don't reach COPY.

Issue #416, opened 2026-09-07
Flyway can't migrate against fakesnow in server mode


anjeongkyun:
Flyway is a common way to manage schema in JVM projects, and it can't currently run against fakesnow in server mode. I traced what stops it, in the order it hits them. Each is small on its own — filing them together since they only make sense as a chain.

Setup: fakesnow server mode over JDBC (`snowflake-jdbc` 3.19.0), `flyway-core` + `flyway-database-snowflake` 10.22.0, two trivial migrations (a `CREATE TABLE` and an `ALTER TABLE ADD COLUMN`).

**1. The login response has no `serverVersion`, so Flyway thinks the server is too old.**

```
Snowflake upgrade required: Snowflake 0.0 is outdated and no longer supported by Flyway.
Flyway currently supports Snowflake 3.0 and newer.
```

The driver reads `data.serverVersion` from the login response (`SessionUtil`) and serves it as `DatabaseMetaData.getDatabaseProductVersion()`. fakesnow doesn't send the field, so `path("serverVersion").asText()` gives `""`:

```
productVersion   = ''
CURRENT_VERSION() = '0.0.0'
```

Note Flyway gates on `getDatabaseProductVersion()`, not `CURRENT_VERSION()` — so this is about the login response specifically.

The value is your call, which is why I didn't send a PR for this one. `transforms.current_version` deliberately returns `0.0.0` ("Needed by sqlalchemy"), but `0.0.0` is also exactly what Flyway rejects. Real Snowflake reports `10.31.103` from `CURRENT_VERSION()` right now. So there's a choice between keeping the two consistent at `0.0.0` (Flyway stays blocked) and reporting something clients accept. Happy to implement whichever you prefer.

**2. `USE ROLE` returns a 500.** Flyway issues it while setting up the connection. That's #415.

**3. `SHOW SEQUENCES` isn't supported.**

```
'SHOW SEQUENCES IN SCHEMA "SCHEMA1"' ProgrammingError 002003 (42S02):
Catalog Error: Table with name SEQUENCES does not exist!
```

Flyway runs this when it inspects the schema for emptiness. This looks like it would follow the existing `show_warehouses`/`show_stages` shape in `transforms/show.py`.

**4. `SHOW PARAMETERS LIKE '...'` isn't parsed.**

```
"SHOW PARAMETERS LIKE 'QUOTED_IDENTIFIERS_IGNORE_CASE'" ProgrammingError 001003 (42000):
[...]

PR #426, opened 2026-09-11
feat: support SHOW PARAMETERS with a basic set

anjeongkyun:
Last of the #416 items.

`SHOW PARAMETERS` wasn't handled:

```
"SHOW PARAMETERS LIKE 'QUOTED_IDENTIFIERS_IGNORE_CASE'" ProgrammingError 001003 (42000):
Parser Error: syntax error at or near "LIKE"
```

Flyway asks for `QUOTED_IDENTIFIERS_IGNORE_CASE` while setting up the connection. It carries on when the query fails, so this one was never blocking, but it does put an error in the log on every connect.

sqlglot parses `SHOW PARAMETERS` as a `Command` rather than a `Show`, so this matches on the command text instead of following the other `show_*` transforms.

It lists two parameters rather than the full Snowflake set: `QUOTED_IDENTIFIERS_IGNORE_CASE` (false, since `alter_session` only accepts false/unset), and `TIMEZONE` (Etc/UTC, what the server reports). Those are the ones fakesnow honours, so the reported values match what the session actually does. Adding more is a one-line entry each.

```
SHOW PARAMETERS                      2 rows
SHOW PARAMETERS IN SESSION           2 rows, scope accepted and ignored
SHOW PARAMETERS LIKE 'quoted%'       1 row, case-insensitive with sql wildcards
SHOW PARAMETERS LIKE 'NOT_A_PARAM'   0 rows
```

With this and #418, the SHOW statements Flyway issues no longer error. It now gets as far as creating its history table and stops on the `TIMESTAMP_LTZ(9)` item I added to #416.

### DIFF
--- a/fakesnow/cursor.py
+++ b/fakesnow/cursor.py
@@ -351,6 +351,7 @@ class FakeSnowflakeCursor:
             .transform(transforms.show_databases)
             .transform(transforms.show_functions)
             .transform(transforms.show_procedures)
+            .transform(lambda e: transforms.show_parameters(e, self._conn._autocommit))
             .transform(transforms.show_warehouses)
             .transform(lambda e: transforms.show_schemas(e, self._conn.database))
             .transform(lambda e: transforms.show_tables_etc(e, self._conn.database, self._conn.schema))
--- a/fakesnow/transforms/__init__.py
+++ b/fakesnow/transforms/__init__.py
@@ -11,6 +11,7 @@ from fakesnow.transforms.show import (
     show_databases as show_databases,
     show_functions as show_functions,
     show_keys as show_keys,
+    show_parameters as show_parameters,
     show_procedures as show_procedures,
     show_schemas as show_schemas,
     show_sequences as show_sequences,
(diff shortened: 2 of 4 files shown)


tekumara:
@anjeongkyun fyi theres some conflicts now

anjeongkyun:
Rebased on main, conflict was just the two tests landing at the same spot in `test_show.py`. Full suite green.

anjeongkyun:
Rebased onto main, no conflicts left. Full suite passes locally apart from `test_cli.py::test_run_server`, which fails the same way on a clean main here, so it looks like my environment rather than the branch.

Worth noting for priority: while checking #416 end to end I found Flyway only warns on the missing `SHOW PARAMETERS` and completes its migrations anyway, so this closes a real gap but is not holding anything up.

tekumara:
@anjeongkyun FYI I've fixed the columns to match real snowflake and removed AUTOCOMMIT because it needs some more work to correctly reflect the value when it changes (plus a few other changes).

PR #428, opened 2026-09-14
fix: PUT with an unquoted file url and honour AUTO_COMPRESS

arnaldojvg:
Alternative to #420, trimmed as discussed there: same unquoted-PUT fix, but only `AUTO_COMPRESS` is honoured — `parallel`, `sourceCompression` and `overwrite` keep their hardcoded defaults (with the existing TODO).

We only need `AUTO_COMPRESS=FALSE`: our job PUTs an already gzipped csv, and with `autoCompress: True` it was gzipped a second time so the staged file no longer matched what `COPY INTO` expected. `put_options` stays generic so other options can be wired later.

If this one is preferred, #420 can be closed and #422 / #424 rebased onto this branch instead.

Part of #409, umbrella #403.


tekumara (approved):
thank you! i've applied some refactoring to make this consistent with the codebase, and split this into independent tests

PR #429, opened 2026-09-17
fix: loading nested Parquet columns with MATCH_BY_COLUMN_NAME

arthur-caix-plana-earth:
## Description

Fixes an issue where `COPY INTO` with `MATCH_BY_COLUMN_NAME` silently loaded semi-structured/variant Parquet columns as `NULL` values instead of their actual data.

### Root Cause
`_get_parquet_column_names` previously filtered `parquet_schema()` using `num_children IS NULL`. This condition excluded any top-level column that was a `struct` or `list` (i.e., nested or variant columns). Because these columns were filtered out, they were never matched against the target table and were skipped during loading.

### Related Link
- https://github.com/tekumara/fakesnow/pull/307

## Key Changes

- Replaced the `parquet_schema()` filtering logic in `_get_parquet_column_names` with `DESCRIBE SELECT * FROM read_parquet(...)`.
- This ensures all top-level columns—including nested and variant types—are correctly retrieved and matched.


tekumara (approved):
thank you!

